Issue #64, opened 2020-11-17
Combine CLI tools


jcapona:
Create a single CLI that wraps calls to the existing commands. 
It should be able to be used directly as a CLI, specifying the desired behavior with arguments.

We need to continue to support the existing CLI, but displaying a warning and pointing the user to the new command

```
$ pitop-config --host
pi-top [4]

$ pt-host
'pt-host' is deprecated, please use 'pitop-config --host'
pi-top [4]
```

PR #74, opened 2020-11-19
Unified CLI (First version)

jcapona:
Closes: #64 

Summary
=======
- Add new cli `pt-config` that replaces old CLI tools.
- Add `PTSocket`class to communicate with `pt-device-manager`. Any rewrite in terms of how CLI communicates with the device manager should be done to this class.
- Refactor CLI commands into classes.
- Add deprecation message to "old" CLI

Usage
=======
```
pi@pi-top:~ $ pt-config -h
usage: pt-config [-h] {brightness,device,host,battery,oled} ...

optional arguments:
  -h, --help            show this help message and exit

Subcommands:
  Set of valid subcommands to use to interface with your pi-top

  {brightness,device,host,battery,oled}
                        valid subcommands
    brightness          communicate and control the device's screen brightness
    device              Get information about device and attached pi-top
                        hardware
    host                Returns the name of the host pi-top device
    battery             Get battery information from a pi-top.
    oled                pi-top OLED quick text
```

Each subcommands help can be checked by running: 
```
pi@pi-top:~ $ pt-config oled -h
usage: pt-config oled [-h] [--timeout TIMEOUT] [--font-size FONT_SIZE] text

positional arguments:
  text                  set the text to write to screen

optional arguments:
  -h, --help            show this help message and exit
  --timeout TIMEOUT, -t TIMEOUT
                        set the timeout in seconds
  --font-size FONT_SIZE, -s FONT_SIZE
                        set the font size
```

Usage remains the same for all cli subcommands:
```
pi@pi-top:~ $ pt-config oled "hey there!" -t 1
```


m-roberts:
I think we should take a look at creating a manpages for:
* `pt-config`
* `pt-config.oled`
* `pt-config.oled`
* `pt-config.oled`
* `pt-config.oled`

docs/cli_tools.rst
```diff
@@ -3,15 +3,41 @@
 ==========================
 
 --------------------
-pt-battery
+pt-config
 --------------------
 
-If the pi-top device has an internal battery, it will report it's status.
+Utility that allows the user to interact with pi-top hardware.
```
m-roberts, line 9:
```suggestion
Utility to interact with pi-top hardware.
```

docs/cli_tools.rst
```diff
@@ -3,15 +3,41 @@
...
+    pt-config [-h] {brightness,device,host,battery,oled} ...
+
+Where:
+
+-h, --help
+    Show a help message and exits
+
+{brightness,device,host,battery,oled}
+    battery:
+        Get battery information from a pi-top.
```
m-roberts, line 22:
```suggestion
        Get battery information from a pi-top
```
For consistency

docs/cli_tools.rst
```diff
@@ -3,15 +3,41 @@
...
+
+    brightness:
+        Communicate and control the device's screen brightness
```
m-roberts, line 25:
```suggestion
        Query and change the device's screen brightness
```

m-roberts:
If this is to be called `pt-config`, I am not sure that the "quick-draw text to OLED" part matches the rest of the functionality, in that it is more for producing an output rather than configuring the system or checking the state, as the rest of this utility does. 

* We could rename it from `pt-config` to `pi-top` as a top level program name, as it suggests that it can be used for configuration and/or handling some of the basic API functionality without needing to write any Python.
* We could keep `pt-config` as a pi-top alternative to `raspi-config`, _specific for configuration_, and then use `pi-top` for handling some of the basic API functionality without needing to write any Python.

m-roberts (changes requested):
Definitely quite a bit still to sort out here

m-roberts:
After discussion with @angusjfw, we have agreed that we should use `pitop`, `pi-top` or `pt` (TBD).

m-roberts:
More work needs to be put into making this CLI make sense, but this is a good start. I will try and develop a proposal for the CLI's API for a future PR.

m-roberts (approved):

PR #75, opened 2020-11-20
pi-top [4] Miniscreen Rework

jcapona:
## Moving modules
- Create new `miniscreen` module, that abstracts pi-top [4]'s OLED screen and buttons.
- Move & rename module `pitop.case_buttons` to `miniscreen.buttons`
- Move module `pitop.oled` to `pitop.miniscreen.oled`

## Renaming variables, classes, files
- Rename PTOLEDDisplay to OLEDDisplay
- Rename `ptoled` variable in examples to `oled_display`
- Rename `PTOLEDManagerException` to `MiniScreenOLEDManagerException`
- Rename `test_ptbuttons.py` to `test_miniscreen_buttons.py`
- Rename `test_ptoled.py` to `test_miniscreen_oled.py`

## Other
- Move example that uses OLED and buttons into separate directory


docs/api_miniscreen.rst
```diff
@@ -0,0 +1,14 @@
+======================================
+API - pi-top [4] Miniscreen
+======================================
+
+pi-top [4]'s case includes a miniscreen module, composed by a OLED screen and 4 buttons.
```
m-roberts, line 5: Not sure what this should read, but this could be clearer/better

docs/api_miniscreen.rst
```diff
@@ -0,0 +1,14 @@
...
+On pi-topOS, the OLED and buttons display an interactive menu, with useful information about
```
m-roberts, line 6: I think we should:
* Mention the package here, so that people can look into it themselves
* Explain how the OLED and button APIs work in parallel with the system menu
* Explain how users can return to the main OLED screen from their code

m-roberts: This appears to be captured in the OLED section, but the same goes for the buttons, so we should probably include a high level note about this here.

.deprecated/ptoled/__init__.py
```diff
@@ -1,4 +1,4 @@
-from pitop.oled import (
+from pitop.miniscreen.oled import (
     PTOLEDDisplay,
```
m-roberts, line 2: I think we need to rename `PTOLEDDisplay`, given this new context

docs/api_miniscreen.rst
m-roberts: Also add to FAQs

docs/api_miniscreen/api_oled.rst
```diff
@@ -17,15 +17,15 @@ until your program exits, at which point the system menu is automatically restor
 yourself with a method of being able to exit, it is recommended that you check for a press event on the
```
m-roberts, line 17: This is a miniscreen point now, not OLED specific as it uses the buttons

.deprecated/ptbuttons/__init__.py
```diff
@@ -1,9 +1,9 @@
...
-from pitop.case_buttons import SelectButton as PTSelectButton  # noqa: F401
-from pitop.case_buttons import CancelButton as PTCancelButton  # noqa: F401
-print("Note, Use of the 'ptbuttons' package is now deprecated. Please use 'pitop.case_buttons' instead.")
+from pitop.miniscreen.buttons import Button as PTButton  # noqa: F401
+from pitop.miniscreen.buttons import Buttons as PTButtons  # noqa: F401
+from pitop.miniscreen.buttons import UpButton as PTUpButton  # noqa: F401
+from pitop.miniscreen.buttons import DownButton as PTDownButton  # noqa: F401
+from pitop.miniscreen.buttons import SelectButton as PTSelectButton  # noqa: F401
+from pitop.miniscreen.buttons import CancelButton as PTCancelButton  # noqa: F401
+print("Note, Use of the 'ptbuttons' package is now deprecated. Please use 'pitop.miniscreen.buttons' instead.")
```
angusjfw, line 7:
```suggestion
print("Note: Use of the 'ptbuttons' package is now deprecated. Please use 'pitop.miniscreen.buttons' instead.")
```

angusjfw (approved):

PR #80, opened 2020-11-23
Create test workflow

m-roberts:
Also uses new base image to speed up package building


olivierwilkinson:
Where are we publishing to PyPI now?

olivierwilkinson (approved):
Approved assuming that publishing to PyPI will be added in it's own pipeline in future. 😉

PR #81, opened 2020-11-24
Patch changelog in Debian Docker container

### LINKED ISSUE #46: Move changelog editing to GitHub Action in a Debian Docker image
Currently this is done in a Ubuntu machine, which is causing problems with versioning. Moving this into a Docker image that is running Debian will ensure that we get the functionality that we want.

m-roberts:
Closes #46

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,9 @@ jobs:
         run: |
           docker create --name py-pitop-chnglg --volume ${{ github.workspace }}:${{ github.workspace }} --workdir ${{ github.workspace }}/ --tty pitop/gbp-dch-gen:latest sleep inf
           docker start py-pitop-chnglg
-          docker exec py-pitop-chnglg gbp dch --ignore-branch --auto --snapshot --distribution=stable --force-distribution
+          docker exec py-pitop-chnglg git config --global user.name "GitHub Build Action"
+          docker exec py-pitop-chnglg git config --global user.email "https://github.com/pi-top/pi-top-Python-SDK"
+          docker exec py-pitop-chnglg gbp dch --auto --snapshot --git-author --ignore-branch
 
       - name: Show updated changelog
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -15,17 +15,15 @@ jobs:
...
-      #   run: |
-      #     gbp dch --ignore-branch --auto --snapshot --distribution=stable --force-distribution
+      - name: Patch changelog (in Debian Docker container)
+        if: github.event_name != 'release'
+        # '--ignore-branch' for testing on feature branch
+        run: |
+          docker create --name py-pitop-chnglg --volume ${{ github.workspace }}:${{ github.workspace }} --workdir ${{ github.workspace }}/ --tty pitop/gbp-dch-gen:latest sleep inf
+          docker start py-pitop-chnglg
+          docker exec py-pitop-chnglg git config --global user.name "GitHub Build Action"
+          docker exec py-pitop-chnglg git config --global user.email "https://github.com/pi-top/pi-top-Python-SDK"
```
angusjfw, line 25: maybe better to use a valid email here like the deb-maintainers one

angusjfw (approved):

debian/changelog
```diff
@@ -1,4 +1,4 @@
-py-pitop (1:0.1.0~prerelease) buster; urgency=medium
+py-pitop (1:0.1.0) UNRELEASED; urgency=medium
```
angusjfw, line 1: jenkins pipeline has an error because this doesn't include 'buster'

m-roberts: Will fix later - this is the correct thing we want as part of a better release process

.github/workflows/build.yml
m-roberts: This is only for preview builds - when we cut a release, we will use a proper email address. Would rather not include group email addresses in a public repo right now

PR #113, opened 2020-12-03
Update motor control registers to match expansion plate labels

duwudi:
Motor labels are M0 to M3, not M1 to M4

Given example code uses M3 so this change still compatible with that


angusjfw:
👍

PR #118, opened 2020-12-07
Update servo motor classes

duwudi:
# Other
* Use f-strings
* Flake8 uses per-file ignores
* Ignore RPi.GPIO RuntimeError when running tests
* Update package to specify 3.7 and above

# Servo

### Latest firmware/hardware spec
- update to be compatible with V1.1 register definition [here] (https://docs.google.com/spreadsheets/d/1bT4LQ4JUOiaeiR4TV7b4TyFxOup9CaKSPcN7KA-uVyI/edit#gid=1254393379) (_This change is not backwards compatible, all expansion plate firmware will need to be updated_)
- There is no mode 0 or mode 1 for servo control anymore, only one mode is required for both speed and angle settings
- Changed servo port numbering to match expansion plate labels (S0-S3) 

### New functionality
- New function for returning **current** speed and angle values which the new firmware makes available (needed for PID control of various planned projects). Deleted the old functions for "get_target_speed" and "get_target_angle" - the user's code already knows what they set anyway so not necessary
- Addition of "acceleration mode" to be either default acceleration (constant speed) or "smoothed" acceleration which makes the servo movement more suitable for humanoid-like movements
- Deleted some test methods but not updated them fully, that will have to be done later


pitop/pma/servo_controller.py
```diff
@@ -87,36 +90,13 @@ def set_target_angle(self, angle: Union[int, float], speed: Union[int, float] =
 
         mapped_speed = int(round(speed * 10))
 
-        self.set_control_mode(ServoControlModes.MODE_1)
+        # self._mcu_device.write_byte(0x50, 1)
```
m-roberts, line 93: Why is this here?

pitop/pma/servo_motor.py
```diff
@@ -66,26 +66,19 @@ def get_angle_range(self):
 
         return (self.min_angle, self.max_angle)
 
-    def get_target_angle(self):
-        """
-        Return the current angle of the servo motor horn.
+    def get_current_angle_and_speed(self):
```
m-roberts, line 69: Not sure about this. We want to use parameters as our means of getting the data (as opposed to a `get_` function). It seems to me that angle and speed should be retrieved separately, or it should return a higher level class with angle and speed properties

pitop/pma/servo_motor.py
```diff
@@ -118,16 +111,6 @@ def set_target_angle(self, angle, speed=50.0):
         self._controller.set_target_angle(angle + self._zero_point, speed)
         self.__has_set_angle = True
 
-    def get_target_speed(self):
```
m-roberts, line 121: Why is this function deleted? I think that `target_angle` should be a property with a getter and setter

m-roberts (commented):
Not sure about this PR. Mostly seems useful and correct, but brings up some questions that need to be answered!

pitop/pma/servo_motor.py
duwudi: We can keep it as a class variable if you want to leave it here, but we won't be reading that value from the MCU anymore. I just assumed that the user code would have access to their own target value that they set

pitop/pma/servo_controller.py
duwudi: whoops

pitop/pma/servo_motor.py
```diff
@@ -46,16 +50,16 @@ def zero_point(self):
...
-            raise ValueError(f"Value must be from {hardware_min_angle} to {hardware_max_angle} degrees (inclusive)")
+        if not (self._HARDWARE_MIN_ANGLE <= zero_position <= self._HARDWARE_MAX_ANGLE):
+            raise ValueError(f"Value must be from {self._HARDWARE_MIN_ANGLE} to {self._HARDWARE_MAX_ANGLE} degrees "
+                             f"(inclusive)")
 
         self._zero_point = zero_position
-        self.min_angle = hardware_min_angle - self._zero_point
-        self.max_angle = hardware_max_angle - self._zero_point
+        self._min_angle = self._min_angle - self._zero_point
+        self._max_angle = self._max_angle - self._zero_point
```
jcapona, line 59: I think we'll lose reference of `_min_angle` and `_max_angle` in here by using the current angles to calculate the new ones...

thinking about it, maybe there are 2 approaches for zero-point setting:
- referencing the "hardware" zero, meaning, the argument `zero_position` refers to the hardware angle, where for example, 0 will always be 0 and it won't change over time
- referencing the latest known zero point setting (and valid angle range), which might differ from the hardware zero (and the original hardware min-max angles). In this approach, the `zero_position` argument might be outside of the hardware angle range...

As this function is now, I think we're mixing these to approaches, since we check against hardware angles at first, but the use the current min-max angles to calculate the new range... I think we should decide on the approach, and document it, since it should be clear to the user!

jcapona (changes requested):
just a comment on something I think we need to clear up before merging this!

pitop/pma/servo_motor.py
duwudi: Good spot, that would've been a serious bug!

pitop/pma/servo_motor.py
duwudi: I've updated everything that can be a property to be one. The get_current_angle_and_speed is useful for people who want both and since they come back from the MCU as one piece of data we should keep it here as getting the data twice using two methods (current_angle and current_speed) would be an double the data overhead (and they'd be out of sync)

duwudi: The only thing that is bothering me at the moment is we have `set_target_angle` as a normal method and we have `target_speed` as a property method - but since set_target_angle requires two arguments I'm not sure how to turn that into a property. We could use a list and have:

`servo.target_angle = [90, 50]` 

but then we force them to have to set a speed at the same time without a default value which I think is maybe worse

Suggestion: we could create a `target_angle` method as a property function that automatically calls a method `set_target_angle_and_speed` with a default speed. Then they have flexibility - I'll do that and we can see how it feels

m-roberts: Good that this is using properties. Can we create a new class that has angle and speed as properties, that this returns an instance of?

duwudi: We could do something like this: https://github.com/pi-top/pi-top-Robotics-Kit-Python-Examples/blob/main/ptrobot/common/data_structures/geometry_msgs.py

But it might get confusing to people as they'll be writing something like `servo.angular.speed` and `servo.angular.angle` to get the values, which will look really similar to `servo.current_angle`

m-roberts: What do you think about this?
```
# Can reflect current or target state
class ServoMotorState:
    def __init__(self, angle, speed):
        self.angle = angle
        self.speed = speed


s = ServoMotor("S1")

# Get
current_state = s.state
angle = current_state.angle
speed = current_state.speed

target_state = s.target_state
angle = target_state.angle
speed = target_state.speed


# Set target speed
s.target_speed = 5

# @target_speed.setter
# def target_speed(self, new_speed):
#     target = self.target_state
#     target.speed = new_speed
#     self.target_state = target


# Set target angle
s.target_angle = 75

# @target_angle.setter
# def target_angle(self, new_angle):
#     target = self.target_state
#     target.angle = new_angle
#     self.target_state = target


# Set target state
target_speed = 5
target_angle = 75

s.target_state = ServoMotorState(target_speed, target_angle)
```

jcapona (dismissed):

jcapona (approved):

PR #119, opened 2020-12-07
Give an initial IMU implementation

duwudi:
Looking for feedback on the general construction of these libs.

I'm considering to use the weakref lib for the encoder and servo motor libraries too so that the robot stops as soon as the code exits (rather than waiting until the heartbeat times out which is a really bad UX)


m-roberts:
Converted to draft PR

examples/pma/imu.py
```diff
@@ -0,0 +1,14 @@
+from pitop.pma import Imu
```
m-roberts, line 1: Would prefer to rename the examples to explain what they are - `imu.py` and `imu2.py` are not very descriptive

examples/pma/imu2.py
```diff
@@ -0,0 +1,23 @@
...
+    if dt > (1 / 10.0):
+        # x, y, z = self.imu_controller.gyroscope_raw
+        gyro = imu.gyroscope
+        print("gyro: {x}, {y}, {z}".format(**gyro))
+        # if axis == 'x':
+        #     degrees_rotated += x * dt
+        # elif axis == 'y':
+        #     degrees_rotated += y * dt
+        # elif axis == 'z':
+        #     degrees_rotated += z * dt
```
m-roberts, line 20: Do we want this?

examples/pma/magnetometer_calibration.py
```diff
@@ -0,0 +1,9 @@
+from pitop.pma.common import ImuCalibration
```
m-roberts, line 1: `imu_` prefix?

pitop/pma/__init__.py
```diff
@@ -11,6 +11,7 @@
 from .joystick import Joystick  # noqa: F401
 from .encoder_motor import EncoderMotor  # noqa: F401
 from .servo_motor import ServoMotor  # noqa: F401
+from .imu import Imu
```
m-roberts, line 14:
```suggestion
from .imu import Imu  # noqa: F401
```

pitop/pma/common/__init__.py
```diff
@@ -1 +1,2 @@
 from .utils import get_pin_for_port, type_check  # noqa: F401
+from .imu_calibration import ImuCalibration
```
m-roberts, line 2:
```suggestion
from .imu_calibration import ImuCalibration  # noqa: F401
```

m-roberts:
Before we merge this in, we need to define the library as 3.7+ due to the use of `dataclass` (which is pretty cool)

m-roberts (approved):

PR #121, opened 2020-12-08
Add weakref cleanup process

duwudi:
Tested and motors stop as soon as the code is exited

### DIFF
--- a/pitop/pma/encoder_motor.py
+++ b/pitop/pma/encoder_motor.py
@@ -1,5 +1,6 @@
 import time
 from math import floor, pi
+import weakref
 
 from .parameters import BrakingType, ForwardDirection, Direction
 from .encoder_motor_controller import EncoderMotorController
@@ -55,6 +56,8 @@ class EncoderMotor:
 
         self._prev_time_dist_cnt = time.time()
         self._previous_reading_odometer = 0
+
+        weakref.finalize(self._motor_core, self._motor_core.cleanup)
 
     @property
     def forward_direction(self):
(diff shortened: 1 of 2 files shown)


pitop/pma/encoder_motor.py
```diff
@@ -56,6 +57,8 @@ def __init__(self, port_name, forward_direction, braking_type=BrakingType.COAST,
         self._prev_time_dist_cnt = time.time()
         self._previous_reading_odometer = 0
 
+        weakref.finalize(self._motor_core, self._motor_core.cleanup)
```
m-roberts, line 60: This looks like an `atexit` style function - is it possible to change the syntax to do the same thing, and maintain consistency?

duwudi: I'm not sure, I got this suggestion from [here ](https://stackoverflow.com/a/42907819/12423177) - I'm not that familiar with this stuff but it seems from other sources too (like [this](https://stackoverflow.com/questions/51513386/how-to-ensure-the-del-function-is-called-on-a-python-class-as-is-commonly-b)) that I would need to use `with` if I wanted to use a __del__ method so if someone used the controller without that then they'd not be guaranteed a cleanup. Is that what you mean? Or can atexit be used as a drop in replacement for the weakref.finalize command I have here?

duwudi: [this answer](https://stackoverflow.com/a/51513387/12423177) using atexit was downvoted for some garbage collection stuff, maybe you'll understand it more

m-roberts:
Closed this in favour of https://github.com/pi-top/pi-top-Python-SDK/commit/5209ecbbec41677095653c0a65696719fdfe1e23#diff-6fa7be83bfd19bbc3d95468b31b5569d78316cc5e764997a207784790b37ab2bR60

PR #124, opened 2020-12-08
Add Development section to readme

angusjfw:


README.rst
```diff
@@ -126,6 +125,44 @@ Comprehensive documentation is available here_.
...
+
+.. code-block:: bash
+
+    python3 -m pytest
+
+And you can build the docs locally by running:
+
+.. code-block:: bash
+
+    PYTHONPATH=. sphinx-build -W -v -bhtml docs/ build/html
```
jcapona, line 159: we should mention that `gpiozero` is a requirement to build the docs! Check this: https://github.com/pi-top/pi-top-Python-SDK/pull/126/commits/a260e817290e2d170d4884fd8a9a015f771cbbef ... If it's not on the system, the class docs won't be generated!

angusjfw: Ah, interesting. But `gpiozero` would be installed by the above `pip3 install -e ".[doc,test]"` so I don't think we need to mention it explicitly.

jcapona: great!

jcapona (approved):

PR #126, opened 2020-12-08
Add overview docs; refactor PMA docs

### LINKED ISSUE #68: Reorganize PMA docs
Improve organization of PMA components, it should be easier to get to a component documentation from the left-side menu.

We should:
- Stop using Input/Output as main categories.
- Mention in each component what kind of device it is (I/O), if it's analog or digital, and where it should be connected.

### LINKED ISSUE #69: Add use cases to "getting started" page
In Getting Started, add new sections along the lines of the following use cases:
- Using the pi-top itself (e.g. buttons, OLED, reading battery, etc.)
- Doing physical computing with pi-top [4] (PMA)
- Using peripherals (e.g. pi-topPULSE/pi-topPROTO+)
- Making use of additional helper classes (e.g. keyboard, speech)

Use cases should have at least one associated recipe, and muchos hyperlinkos

jcapona:
Closes #68
Closes #69

### Summary

- Reorganize PMA docs
- Fix readthedocs builds
- Add overview page


docs/api_pma/components.rst
```diff
@@ -0,0 +1,116 @@
+
+Button
+-------------------------------
+
+.. image:: ../_static/pma/foundation_kit/components/button.jpg
+
+A Button is a digital input device; it generates a boolean value whenever it's pressed or released.
```
m-roberts, line 7:
```suggestion
A Button is a binary digital input device; this means that it has 2 states - pressed or released. This class provides an easy way to query the state of the device as well as perform event-based actions when it changes state.
```

m-roberts: What about if we create a page that explains the ports for each type of device, and link to that from "digital input device"? We can remove the need to explain the available ports for each component; or perhaps we can `.. include` it from somewhere so that it isn't copy/pasted everywhere?

m-roberts (approved):

m-roberts (approved):

PR #135, opened 2020-12-14
Set OLED cli force argument default value to False

jcapona:


debian/changelog
```diff
@@ -1,3 +1,9 @@
+py-pitop-sdk (0.2.6) buster; urgency=medium
+
+  * Update OLED cli force argument default value to False
```
m-roberts, line 3:
```suggestion
  * Set default OLED CLI force argument to False
```

m-roberts (approved):

PR #157, opened 2021-01-06
Adaptive OLED SPI (+ use latest common lib)

jcapona:
* Use `parameters` property, not function
* Remove unused `zmq` dependency (handled in common lib)
* OLED uses controller class
* Add additional top-level OLED properties for size and mode
* Add additional top-level OLED functions for contrast
* Canvas and FPS regulators are now private
* Add `pi-top oled spi` command for changing SPI bus


pitop/miniscreen/oled/core/controls.py
```diff
@@ -53,7 +56,7 @@ def __setup_device():
             port=spi_port,
             device=spi_device,
             bus_speed_hz=spi_bus_speed_hz,
-            cs_high=spi_cs_high,
+            # cs_high=spi_cs_high,
```
m-roberts, line 59: Let's move to latest Luma so that cs_high is not a concern any more.

pitop/miniscreen/oled/core/controls.py
```diff
@@ -6,6 +6,9 @@
 from luma.core.interface.serial import spi
 from luma.oled.device import sh1106
 
+from pitop.system.daemon import get_oled_spi_port
```
m-roberts, line 9: Move the request client logic into this file. You may need to turn this file into a class of sorts to facilitate this.
This is something that will need to happen sooner or later anyway, so probably makes sense now.

m-roberts: Use battery as a guide

pitop/miniscreen/oled/oled.py
```diff
@@ -53,16 +53,20 @@ def __init__(self):
 
     def __setup_subscribe_client(self):
         def on_control_changed(parameters):
-            controller = int(parameters[0])
+            controller = int(parameters()[0])
```
m-roberts, line 56: I still worry about needing to invoke 'parameters'. Can we change this now before it's all across our codebase?

pitop/miniscreen/oled/oled.py
```diff
@@ -53,16 +53,20 @@ def __init__(self):
...
             else:
                 self.__ptdm_subscribe_client.invoke_callback_func_if_exists(self.when_hub_takes_control)
 
+        def on_spi_port_changed(parameters):
+            self.__ptdm_subscribe_client.invoke_callback_func_if_exists(self.reset)
+
         self.__ptdm_subscribe_client = PTDMSubscribeClient()
         self.__ptdm_subscribe_client.initialise({
             Message.PUB_OLED_CONTROL_CHANGED: on_control_changed,
+            Message.PUB_OLED_SPI_BUS_CHANGED: on_spi_port_changed,
```
m-roberts, line 69: lambda function would be better here - see buttons for reference

pitop/system/daemon.py
```diff
@@ -8,3 +8,19 @@ def ping():
         response = request_client.send_message(message)
 
     return response.parameters[0]
+
+
+def set_oled_spi_port(port):
```
m-roberts, line 13: Move to OLED class

m-roberts: Set as property (i.e. `oled.spi_bus(), oled.spi_bus(1)`

m-roberts:
Also the CLI needs to have: `pi-top oled spi 1`

m-roberts:
Marked as ready for review, as this should work with PTDM once we have addressed this in device manager:
```
>>> from pitop.miniscreen.oled.core.device_controller import OledDeviceController
>>> o = OledDeviceController()
>>> o.spi_bus
1
>>> o.spi_bus = 0
  File "/usr/lib/python3/dist-packages/pitop/miniscreen/oled/core/device_controller.py", line 116, in spi_bus
    request_client.send_message(message)
  File "/usr/lib/python3/dist-packages/pitopcommon/ptdm.py", line 372, in send_message
    response_string = self.__zmq_socket.recv_string()
  File "/usr/lib/python3/dist-packages/zmq/sugar/socket.py", line 584, in recv_string
    msg = self.recv(flags=flags)
  File "zmq/backend/cython/socket.pyx", line 788, in zmq.backend.cython.socket.Socket.recv
  File "zmq/backend/cython/socket.pyx", line 824, in zmq.backend.cython.socket.Socket.recv
  File "zmq/backend/cython/socket.pyx", line 191, in zmq.backend.cython.socket._recv_copy
  File "zmq/backend/cython/socket.pyx", line 186, in zmq.backend.cython.socket._recv_copy
  File "zmq/backend/cython/checkrc.pxd", line 19, in zmq.backend.cython.checkrc._check_rc
zmq.error.Again: Resource temporarily unavailable
```

which seems to be a timing issue

pitop/battery/battery.py
```diff
@@ -22,27 +22,25 @@ def __init__(self):
...
             else:
-                self.__ptdm_subscribe_client.invoke_callback_func_if_exists(self.when_charging)
+                if callable(self.when_charging):
+                    self.when_charging()
 
         self.__ptdm_subscribe_client = PTDMSubscribeClient()
         self.__ptdm_subscribe_client.initialise({
-            Message.PUB_LOW_BATTERY_WARNING: on_low_battery,
-            Message.PUB_CRITICAL_BATTERY_WARNING: on_critical_battery,
+            Message.PUB_LOW_BATTERY_WARNING: lambda: self.when_low,
```
m-roberts, line 42: What happens if this is passed in as `None`, but is later set? Is the reference updated?

m-roberts (approved):

PR #172, opened 2021-01-20
OLED refactor

m-roberts:
### Changes
OLED class
* `reset()` now clears the screen, as a user would reasonably expect
* Added a `refresh()` function to do what `reset()` was previously doing
* `oled.draw_` deprecated; `oled.display_` now used
* Revert previous OLED refactor - multiple canvases in OLED, additional functions, etc.
* Add `prepare_image` method to OLED for processing its size and mode
* Add positional properties to OLED (e.g. `top_left`, etc. and use directly)
* Pass 'redraw last image' to OLED device controller to clarify scope of 'reset' during SPI bus change

Canvas class
* Canvas object, whilst now technically deprecated, has better scope management with its new constructor taking `mode` and `size` rather than an OLED device
* Only perform image processing for OLED if size or mode are not correct (performance improvement)
* Drop canvas save function in favour of sample code for saving an OLED image

Docs
* Dropped canvas from docs, as this is no longer the intended way to use the OLED

Examples
* Drop unused `set_max_fps` in examples
* Handle images directly, no canvas

CLI
* Add OLED capture CLI

Other
* FPS regulator back in its own file
* OLED's `LOCK_FILE_PATH` and `OledDeviceController ` are now private

### Later
* Add example of writing text on top of a camera image on OLED, using `process_image` before adding text to show how to handle scaling an image before annotating
* Update docs to communicate high-level OLED handling (PIL images, links to how to use, etc.)


pitopcli/oled.py
```diff
@@ -28,9 +27,19 @@ def is_animated(image):
...
+            elif isdir(arg_path):
+                file_path = join(arg_path, f"output_{strftime('%Y-%m-%d-%H-%M-%S')}")
+            else:
+                raise Exception(f"Invalid path: {arg_path}")
+            return file_path
+
         try:
-            if self.args.oled_subcommand == "draw":
-                # Do take control of OLED to draw
+            if self.args.oled_subcommand == "display":
```
jcapona, line 41: update the docs!

m-roberts: Thanks!

PR #197, opened 2021-02-01
Improve image conversions

### LINKED ISSUE #183: Improve image handling (RGB-BGR/format conversion)
Handle images in a better way. Currently, PIL/OpenCV image conversion automatically handles RGB to BGR conversion, but can be mishandled to produce images which do not have correct channel ordering.

It seems sensible to me to move format checking (PIL/OpenCV) to ImageFunctions, in a more robust way, such as:
```
def convert(image, format="PIL", rgb_to_bgr=None):
    if rgb_to_bgr is None:
        # auto detect
         if typeof(image) == PIL and format == "PIL":
             rgb_to_bgr = False
         elif typeof(image) == array and format == "OpenCV":
             rgb_to_bgr = False
         else:
             rgb_to_bgr = True
    image_arr = array(image)
    if rgb_to_bgr and not __image_has_3_channels(image_arr):
        print("Unable to perform RGB to BGR conversion - 3 channels areq required.")
    if rgb_to_bgr:
        image = __image_rgb_bgr_convert(image_arr)
    else:
        image = image_arr
    if format == "PIL":
        image = Image.fromarray(image)
    return image
```

Also, it appears that the 1-bit PIL images from the OLED are 2-dimensional, not 3 as with RGB. This means that it currently cannot be converted to JPG correctly when being sent to via Further Link's `send_image` function.

jcapona:
Closes #183

Implement generic function to convert between different image formats (for now, `PIL` <-> `OpenCV`).


debian/source/lintian-overrides
```diff
@@ -0,0 +1 @@
+py-pitop-sdk source: out-of-date-standards-version 4.5.0 (released 2020-01-20) (current is 4.5.1)
```
m-roberts, line 1: Why not just bump it?

docs/recipes_api.rst
```diff
@@ -26,6 +26,11 @@ Robotics Kit: Alex Robot - Line Detection
 
 .. literalinclude:: ../examples/robotics/alex_line_detect.py
 
+Displaying camera stream in pi-top [4]'s miniscreen
+------------------------------------------------------------
```
m-roberts, line 30:
```suggestion
Displaying camera stream in pi-top [4]'s miniscreen
---------------------------------------------------
```

docs/recipes_api.rst
```diff
@@ -26,6 +26,11 @@ Robotics Kit: Alex Robot - Line Detection
...
+
+.. literalinclude:: ../examples/camera/camera_display_in_miniscreen.py
```
m-roberts, line 32: How do we handle the examples that combine things together? Should there be an `examples/recipes` folder?

pitop/core/ImageFunctions.py
```diff
@@ -1,60 +1,42 @@
...
+def convert(image, format="PIL"):
+    try:
+        from cv2 import (
+            cvtColor,
+            COLOR_BGR2RGB,
+            COLOR_RGB2BGR,
+        )
+    except (ImportError, ModuleNotFoundError):
+        raise ModuleNotFoundError(
+            "OpenCV Python library is not installed. You can install it by running 'sudo apt install python3-opencv'.") from None
```
m-roberts, line 20: What about libatlas?

pitop/core/ImageFunctions.py
```diff
@@ -1,60 +1,42 @@
...
+
+    assert isinstance(format, str)
+    format = format.lower()
+    assert format in ("pil", "opencv")
```
m-roberts, line 24: Seems like a nice generic format check function that could be used wherever `format=` is used?

pitop/core/ImageFunctions.py
```diff
@@ -1,60 +1,42 @@
...
+
+    if any([
+        isinstance(image, Image.Image) and format == "pil",
+        isinstance(image, ndarray) and format == "opencv"
+    ]):
+        return image
```
m-roberts, line 30:
```suggestion
    # Image type is already correct - return image
    if any([
        isinstance(image, Image.Image) and format == "pil",
        isinstance(image, ndarray) and format == "opencv"
    ]):
        return image
```

pitop/core/ImageFunctions.py
```diff
@@ -1,60 +1,42 @@
...
+    elif isinstance(image, Image.Image) and format == "opencv":
+        cv_image = asarray(image)
```
m-roberts, line 32:
```suggestion
        # Convert PIL to OpenCV
        cv_image = asarray(image)
```

pitop/core/ImageFunctions.py
```diff
@@ -1,60 +1,42 @@
...
+        if image.mode == "RGB":
+            cv_image = cvtColor(cv_image, COLOR_RGB2BGR)
+        return cv_image
+    elif isinstance(image, ndarray) and format == "pil":
+        if len(image.shape) > 2 and image.shape[2] == 3:
```
m-roberts, line 37: Can we explain this a bit?

m-roberts, line 37:
```suggestion
        # Convert OpenCV to PIL
        if len(image.shape) > 2 and image.shape[2] == 3:
```

m-roberts (changes requested):

debian/source/lintian-overrides
m-roberts, line 1:
```suggestion
# Jenkins fails with 4.5.1, so need to ignore the Lintian info tag until this is resolved
py-pitop-sdk source: out-of-date-standards-version 4.5.0 (released 2020-01-20) (current is 4.5.1)
```

m-roberts (approved):

PR #203, opened 2021-02-03
Example: use bluedot to control Alex

jcapona:


docs/recipes_api.rst
```diff
@@ -26,6 +26,22 @@ Robotics Kit: Alex Robot - Line Detection
 
 .. literalinclude:: ../examples/robotics/alex_line_detect.py
 
+Robotics Kit: Alex Robot - Control using Bluedot
+-------------------------------------------------
+
+`BlueDot <https://bluedot.readthedocs.io/en/latest/>`_ is a python library that allows you to control Raspberry Pi projects remotely with a device that runs Android. In this case, we'll use it as a Joystick to control Alex.
+
+Please follow these instructions to setup your pi-top with BlueDot:
```
m-roberts, line 39: I think we should link to https://bluedot.readthedocs.io/en/latest/gettingstarted.html for this, rather than a description. I would say after you link that "Code that uses BlueDot will pause until you select your device from the device list in the  BlueDot app." or something.

m-roberts (approved):

PR #212, opened 2021-02-10
Add giphy miniscreen startup example

duwudi:
Not sure on the name of this file, feel free to suggest alternatives


m-roberts:
* Needs to be included in the docs
* Should include explanation of how it works outside of the code (e.g. requires the latest version of `pt-sys-oled`)

jcapona (approved):

PR #254, opened 2021-02-25
[WIP] Improve auto-wrapping text

### LINKED ISSUE #253: Improve auto-wrapping text
See [here](https://forum.pi-top.com/t/miniscreen-multiline-text/562/2) for discussion of the problem.

Thinking about this, `xy` actually makes word wrap a bit awkward...

[This](https://docs.wand-py.org/en/0.6.2/guide/draw.html#word-wrapping) does it differently - it automatically sets the text size and wrapping to fit in a bounding box. This is much more intuitive, I think>

[This implementation](https://gist.github.com/pojda/8bf989a0556845aaf4662cd34f21d269) also does something similar.

m-roberts:
Closes #253

This is a draft PR, and is not final.
Still needs a clear strategy for how to find the maximum font size so that the whole text fits into the box, considering height and width (w/word wrapping)
Note: should have a minimum font size

TODOs:
https://github.com/pi-top/pi-top-Python-SDK/pull/254/files#diff-8ff4c1bbe8c2ef3c263d407574ea58158611e8d954011854873177db4d3346e4R20

https://github.com/pi-top/pi-top-Python-SDK/pull/254/files#diff-8ff4c1bbe8c2ef3c263d407574ea58158611e8d954011854873177db4d3346e4R62


m-roberts:
@jcapona - what do you think?

m-roberts:
This PR is quite old now. It's probably best used as guidance for how to write it properly rather than as the basis for an actual code submission

PR #256, opened 2021-02-26
Initial Preview: Robot controller via web input

### LINKED ISSUE #204: Add virtual controller via web frontend to Robotics Kit
Robotics Kit needs an easy way for people to be able to control!

m-roberts:
Closes #204

Introducing this via `pitop.labs`, so as to allow us to develop this without having to commit to one particular implementation without better understanding our requirements.

# Usage

```
from pitop import Pitop, DriveController, PanTiltController, Camera
from pitop.labs.webserver import run_webserver

# Create components
robot = Pitop()
drive_controller = DriveController()
pan_tilt = PanTiltController()
camera = Camera()

# Add components to the robot object
robot.add_component(drive_controller)
robot.add_component(pan_tilt)
robot.add_component(camera)

# Run webserver
run_webserver(robot_instance=robot,
              port=8070,
              serve_forever=True)
```


angusjfw:
> @jcapona @m-roberts even by reducing the quality of the jpeg like this it's still pretty much unusable on WiFi

@duwudi I think this is fixed by PR #262

pitop/labs/webserver/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+
+    with app.app_context():
+        from . import routes  # noqa: F401
+        app.config['robot'] = robot_instance
+        robot_instance.camera.on_frame = routes.handle_frame
+        return app
+
+
+def run_webserver(robot_instance, port=8070, serve_forever=True):
+    print(f"\n\nOpen a new tab in your browser and go to http://127.0.0.1:{port}/")
```
jcapona, line 28: get device IP's

m-roberts:
I would prefer it if `nipplejs.js` were minified as part of the build (`debian/rules`), but this is fine

pitop/labs/webserver/routes.py
```diff
@@ -0,0 +1,78 @@
...
+    msg_data = m.get('data', dict()).get('data', dict())
+    robot = app.config['robot']
+
+    if msg_type == 'cmd_vel':
+        linear_speed = msg_data.get("linear", dict()).get("x")
+        angular_speed = msg_data.get("angular", dict()).get("z")
+        robot.drive.robot_move(linear_speed, angular_speed)
+
+    elif msg_type == 'pan_tilt':
+        conversion_factor = 90.0 / 100.0
```
duwudi, line 73: what's the conversion factor for?

pitop/labs/webserver/static/joysticks.js
```diff
@@ -0,0 +1,106 @@
...
+  'linear': {
+    'x': 0.0
+  },
+  'angular': {
+    'z': 0.0
+  }
+};
+
+var pan_tilt_twist = {
+  'angular': {
```
duwudi, line 11: This should be changed to angle

pitop/labs/webserver/static/joysticks.js
```diff
@@ -0,0 +1,106 @@
+var cmd_vel_twist = {
+  'linear': {
+    'x': 0.0
+  },
+  'angular': {
+    'z': 0.0
+  }
+};
+
+var pan_tilt_twist = {
```
duwudi, line 10: This should be changed to pan_tilt_position

pitop/labs/webserver/static/joysticks.js
```diff
@@ -0,0 +1,106 @@
...
+    // console.log(linear)
+    window.command['cmd_vel'](cmd_vel_twist);
+}
+
+function panTiltTwistPublisher(angular_y, angular_z) {
+    if (angular_y !== undefined && angular_z !== undefined) {
+        pan_tilt_twist.angular.y = angular_y;
+        pan_tilt_twist.angular.z = angular_z;
+    } else {
+        pan_tilt_twist.linear.x = 0;
```
duwudi, line 40: there's a copy-paste bug here, pan_tilt_twist doesn't have a linear component. However, the new implementation of setting the angle directly instead of the speed actually works quite well, so see other comments

pitop/labs/webserver/static/joysticks.js
```diff
@@ -0,0 +1,106 @@
...
+            if (publishCmdVelImmediately) {
+              publishCmdVelImmediately = false;
+              cmdVelTwistPublisher(linear, angular);
+              setTimeout(function () {
+                publishCmdVelImmediately = true;
+              }, 50);
+            }
+        } else {
+            let angular_y = -Math.cos(direction / 57.29) * joystick.distance * MAX_SERVO_SPEED / 100.0;
+            let angular_z = Math.sin(direction / 57.29) * joystick.distance * MAX_SERVO_SPEED / 100.0;
```
duwudi, line 77: Change to angle_y and angle_z. Remove MAX_SERVO_SPEED and change to MAX_SERVO_ANGLE instead since this implementation uses position and not angular speed

pitop/labs/webserver/static/joysticks.js
```diff
@@ -0,0 +1,106 @@
...
+        cmd_vel_twist.angular.z = angular;
+    } else {
+        cmd_vel_twist.linear.x = 0;
+        cmd_vel_twist.angular.z = 0;
+    }
+    // console.log(linear)
+    window.command['cmd_vel'](cmd_vel_twist);
+}
+
+function panTiltTwistPublisher(angular_y, angular_z) {
```
duwudi, line 35: change to panTiltPositionPublisher

duwudi:
Really like this, sending the servo angle directly actually works very well - especially since it resets the servos once the user is finished looking which was a problem with the servo speed implementation used before (driving with an offset servo is not easy!). 

The reason we might want to implement a servo speed one in future is to have a "pan-follow" drive mechanism, so when the pan servo moves the chassis would move to follow it - essentially this means the user is looking at where they want to steer towards and in steady-state you still get the zero angle pan. However, this implementation is way simpler and has a great UX so nice work!

duwudi:
> > @jcapona @m-roberts even by reducing the quality of the jpeg like this it's still pretty much unusable on WiFi
> 
> @duwudi I think this is fixed by PR #262

@jcapona I noticed this PR is not the same as what was in #262 - are we sure this is fixed as I still have way too much latency, it's more than what is necessary for true FPV driving (I have to be looking at the robot to drive it without crashing)

jcapona:
> > > @jcapona @m-roberts even by reducing the quality of the jpeg like this it's still pretty much unusable on WiFi
> > 
> > 
> > @duwudi I think this is fixed by PR #262
> 
> @jcapona I noticed this PR is not the same as what was in #262 - are we sure this is fixed as I still have way too much latency, it's more than what is necessary for true FPV driving (I have to be looking at the robot to drive it without crashing)

yeah... I had to revert #262 (cc: @angusjfw ) because the image stream blocks the execution of the app, so joysticks don't work (their commands aren't sent)

angusjfw:
> yeah... I had to revert #262 (cc: @angusjfw ) because the image stream blocks the execution of the app, so joysticks don't work (their commands aren't sent) ... and [adb9455](https://github.com/pi-top/pi-top-Python-SDK/commit/adb9455c9f918f19e55d7b24df3500496325218f) behaves the same way

Right, of course, I've fixed it to make both systems work now. I've been driving it around with this on my phone and it seems pretty good. Camera direction control brings a lot of life to it.

I also removed the `optimize=True` flag when converting the camera frames as I measured this was increasing the latency of the video stream. For live video the lost time optimizing outweighs the benefit of sending less data.

Anyway, we should discuss properly how we want to approach concurrency issues like this one we've found with handling camera and joysticks. The use of gevent in this server might complicate the situation for users and other parts of the sdk as it can be tricky to use alongside threads and asyncio, which are probably what users would prefer to work with and/or learn.

angusjfw (approved):
This seems to work just fine. Given it's in 'labs' and we are expecting to make further improvements I think we should merge it in as a good starting point.

Some of the things I think we'll want to look at in future:
- user customisation of UI, commands & behaviours along the lines of @olivierwilkinson's suggestions (pr to come)
- possibly switching from a gevent based server to asyncio will make the route handlers easier to work with and more transferrable learning for users (and us)
- making the frontend more of a generic interface, for example by naming the websocket messages leftJoystick/rightJoystick and handling twist calculations on the backend, so that they can be more readily customised (for example switched to two stick tank driving). possibly there could be javascript options for a ROS and other backend if that is what the naming (cmd_vel) attempts to target
- resolving other issues with the javascript like sending the stop message three times which I don't think has fixed the issue it addresses.

m-roberts:
Just to add on top of what Angus has put (which I completely agree with), I would also like to think about how we can cover the web stuff with tests, and how we can/if we should combine this with coverage reporting from the Python code.

PR #261, opened 2021-03-05
Track state and recreate Pitop object

jcapona:
# Summary

- Track state of components. What "state" means is defined in each class, using the `own_state` method. This is performed using the `Stateful` mixin.
- Allow component recreation using a dictionary/JSON file & saving a component configuration to file.  This is performed via the `Recreatable` mixin.
- Allow to "attach" components to a main object, and keep track of the state & configuration of it, through the `Componentable` mixin.
- Provide deprecation support to the `AlexRobot` class, by adding a JSON file that is loaded whenever the class is instantiated.

# Code sample

```

>>> from pitop import Pitop, LED
>>> p = Pitop()
>>> led = LED("D2")
>>> p.add_component(led)

# get state as a dict
>>> p.component_state
{'led': {'is_lit': False, 'value': 0}}

# get config as a dict
>>> p.component_config
{'classname': 'Pitop', 'module': 'pitop.system.pitop', 'version': '0.17.0', 'components': {'led': {'port_name': 'D2', 'name': 'led', 'classname': 'LED', 'module': 'pitop.pma.led', 'version': '0.17.0'}}}

# pretty-print methods
>>> p.print_state()
{
    "led": {
        "is_lit": false,
        "value": 0
    }
}
>>> p.print_config()
{
    "classname": "Pitop",
    "module": "pitop.system.pitop",
    "version": "0.17.0",
    "components": {
        "led": {
            "port_name": "D2",
            "name": "led",
            "classname": "LED",
            "module": "pitop.pma.led",
            "version": "0.17.0"
        }
    }
}

# save configuration
>>> p.save_config("/home/pi/my_config.json")

### CODE REVIEW

File: examples/system/pitop_load_configuration.py [hunk 1]
```diff
@@ -0,0 +1,20 @@
+from time import sleep
+
+from pitop import Pitop
+
+
+# Load configuration from a previous session
+pitop = Pitop.from_file("/home/pi/my_custom_config.json")
```
jcapona, line 7: We probably want to provide a real json file, other than `alex.json`, for us to use in here? 
My idea was that in the docs you'll see the "save the configuration" example first and this one next, so it kind of makes sense in that context.

pitop/core/mixins/supports_driving.py
```diff
@@ -0,0 +1,16 @@
...
+class SupportsDriving(DriveController):
+    def __init__(self, motor_left_port, motor_right_port, **kwargs):
+        try:
+            if connected_plate() != FirmwareDeviceID.pt4_expansion_plate:
+                raise Exception
+            DriveController.__init__(self, motor_left_port, motor_right_port)
+            if hasattr(self, "add_component"):
+                self.add_component(self, "drive_controller")
+        except Exception:
+            print("No DriveController")
```
jcapona, line 16: Improve message: `No expansion plate detected: DriveController isn't available` or something like that

pitop/robotics/configurations.py
```diff
@@ -0,0 +1,23 @@
...
+from pathlib import Path
+
+from pitop import Pitop
+
+
+def alex_configuration():
+    return __load_json('alex.json')
+
+
+def AlexRobot():
```
jcapona, line 11: Should we delete this? In the Alex examples we're now using `Pitop.from_dict(alex_configuration())`

pitop/robotics/configurations.py
```diff
@@ -0,0 +1,23 @@
+from json import load
+from pathlib import Path
+
+from pitop import Pitop
+
+
+def alex_configuration():
```
jcapona, line 7: might be better to define this as a variable instead of a function

docs/api_pitop_device.rst
```diff
@@ -32,6 +32,19 @@ Class Reference: Pitop
 
 All pi-tops come with some software-controllable onboard hardware. These sections of the API make it easy to access and change the state of your pi-top hardware.
 
+Using the Pitop object
```
m-roberts, line 35: This file should be a lot clearer. Let's get others to review this.

examples/recipes/alex_line_detect.py
```diff
@@ -1,18 +1,18 @@
-from pitop import AlexRobot
+from pitop import Pitop, alex_configuration
```
m-roberts, line 1: Not sure I like `alex_configuration`. Let's see what others think.

examples/recipes/alex_line_detect.py
```diff
@@ -1,18 +1,18 @@
...
 from pitop.processing.algorithms.line_detect import process_frame_for_line
 
 from signal import pause
 
 
-# Set up robot
-alex = AlexRobot()
+# Load Alex configuration into a Pitop object
+alex = Pitop.from_dict(alex_configuration)
```
m-roberts, line 8: I would suggest that `from_config` is more appropriate here. It may be a dict, but it's a specifically organised dict that represents a pi-top's configuration.

examples/recipes/alex_line_detect.py
```diff
@@ -1,18 +1,18 @@
...
 
 
 # Set up logic based on line detection
 def drive_based_on_frame(frame):
     processed_frame = process_frame_for_line(frame)
     print(f"Target angle: {processed_frame.angle:.2f} deg ", end="\r")
-    alex.target_lock_drive_angle(processed_frame.angle)
+    alex.drive.target_lock_drive_angle(processed_frame.angle)
```
m-roberts, line 15: Not entirely related to this PR, but what is `target_lock_drive_angle`? Can we get some docstrings?

examples/system/pitop_load_configuration.py
```diff
@@ -0,0 +1,20 @@
+from time import sleep
+
+from pitop import Pitop
```
m-roberts, line 3:
```suggestion
from pitop import Pitop
from time import sleep
```

examples/system/pitop_load_configuration.py
```diff
@@ -0,0 +1,19 @@
...
+from time import sleep
+
+
+# Load configuration from a previous session
+pitop = Pitop.from_file("/home/pi/my_custom_config.json")
+
+# Check the loaded configuration
+pitop.component_config
```
m-roberts, line 9: What is this?

examples/system/pitop_load_configuration.py
```diff
@@ -0,0 +1,19 @@
...
+
+# Do something with your device
+pitop.red_led.on()
+pitop.drive.forward(0.5)
+sleep(2)
+pitop.red_led.off()
+pitop.drive.stop()
+
+# Check the state of all the components attached to the Pitop object
+pitop.component_state
```
m-roberts, line 19: I think this needs a `print()` wrapped around it, to be run in a script.

pitop/__init__.py
```diff
@@ -2,29 +2,31 @@
...
     Direction,
-    BrakingType
+    BrakingType,
 )
 
 # # Robotics
-from .robotics.alex_robot import AlexRobot
+from .robotics.configurations import (
+    alex_configuration,
+    AlexRobot,
```
m-roberts, line 28:
```suggestion
    AlexRobot,  # deprecated
```

pitop/camera/camera.py
```diff
@@ -59,6 +64,24 @@ def __init__(self,
...
+            "index": index,
+            "resolution": resolution,
+            "camera_type": camera_type.value if isinstance(camera_type, Enum) else camera_type,
+            "path_to_images": path_to_images,
+            "format": format,
+            "name": self.name,
+        })
+
+    @property
+    def own_state(self):
```
m-roberts, line 79: Not sure that I like this property name

pitop/camera/core/cameras/usb_camera.py
```diff
@@ -9,6 +9,7 @@ def __init__(self, index: int = 0, resolution=None):
         self.index = index
 
         try:
+            self.__camera = None
```
m-roberts, line 12:
```suggestion
        self.index = index
        self.__camera = None

        try:
```

pitop/core/mixins/supports_camera.py
```diff
@@ -0,0 +1,17 @@
+from pitop.core.exceptions import UnavailableComponent
+
+
+class SupportsCamera:
```
m-roberts, line 4: This appears to be unused, including examples

pitop/core/mixins/supports_driving.py
m-roberts: Agreed - all print-outs should be reviewed.

pitop/pma/adc_base.py
```diff
@@ -16,13 +21,23 @@ class ADCBase:
     :param str port_name: The ID for the port to which this component is connected
     """
 
-    def __init__(self, port_name, pin_number=1):
+    def __init__(self, port_name, pin_number=1, name=""):
```
m-roberts, line 24: Should default name be 'unset ADC base' or something?

pitop/pma/adc_base.py
```diff
@@ -57,3 +72,7 @@ def __read_peak(self, channel):
     # read 16 bits register
     def __read_register(self, read_address):
         return self.__adc_device.read_unsigned_word(read_address, little_endian=True)
+
+    @property
+    def value(self):
+        pass
```
m-roberts, line 78: I am guessing that this gets overridden? Should we have `NotImplementedError` here?

pitop/pma/encoder_motor.py
```diff
@@ -62,6 +68,21 @@ def __init__(self,
 
         atexit.register(self.stop)
 
+        Stateful.__init__(self)
+        Recreatable.__init__(self, config_dict={'port_name': port_name, 'name': name})
+        # Added as lambdas since they may change on runtime
```
m-roberts, line 73: This would be better explained in the `add_to_config` function docsting.

pitop/pma/encoder_motor.py
```diff
@@ -62,6 +68,21 @@ def __init__(self,
...
+        self.add_to_config("forward_direction", lambda: self.forward_direction)
+        self.add_to_config("braking_type", lambda: self.braking_type)
+        self.add_to_config("wheel_diameter", lambda: self.wheel_diameter)
+
+    @property
+    def own_state(self):
+        return {
+            'current_rpm': lambda: self.current_rpm,
+            'current_speed': lambda: self.current_speed,
+            'distance': lambda: self.distance,
```
m-roberts, line 83: There's quite a bit more useful information that could be provided here. Braking type? Wheel diameter?

pitop/pma/light_sensor.py
```diff
@@ -15,6 +15,9 @@ class LightSensor(ADCBase):
     :param str port_name: The ID for the port to which this component is connected
     """
 
+    def __init__(self, port_name, pin_number=1, name="light_sensor"):
```
m-roberts, line 18: Is this file missing a state property?

pitop/pma/encoder_motor.py
m-roberts: Or perhaps that is more "config" rather than "state. But I wonder why we wouldn't include that as well...

pitop/robotics/configurations.py
m-roberts: Definitely. Perhaps it could be an enum of stock configurations?

pitop/robotics/configurations.py
m-roberts: Keep it for legacy support. We won't advertise it.

pitop/robotics/drive_controller.py
```diff
@@ -1,56 +1,66 @@
...
+from pitop.pma import (
     EncoderMotor,
     ForwardDirection,
 )
 
-from pitop.system.port_manager import PortManager
-from simple_pid import PID
 
+class DriveController(Stateful, Recreatable):
+    """Abstraction of a vehicle with two wheels connected by an axis, and an
```
m-roberts, line 22: Let's avoid "abstraction of", and just describe straight away. It doesn't add any value. "Represents" would be better.

pitop/robotics/drive_controller.py
```diff
@@ -66,48 +76,107 @@ def __calculate_motor_rpms(self, linear_speed, angular_speed, turn_radius):
...
-        self._right_motor.set_target_rpm(target_rpm=rpm_right,
-                                         total_rotations=rotations*rpm_right/abs(rpm_right))
+        self.left_motor.set_target_rpm(target_rpm=rpm_left,
+                                       total_rotations=rotations*rpm_left/abs(rpm_left))
+        self.right_motor.set_target_rpm(target_rpm=rpm_right,
+                                        total_rotations=rotations*rpm_right/abs(rpm_right))
+        sleep(time_to_take)
 
     def stop(self):
+        """Completely stops the robot."""
```
m-roberts, line 167: Think of function docstrings as finishing the sentence "This function, when called, will..."

* Rotate the robot in place by a given angle and stop ✅
* Completely stops the robot ❌

pitop/robotics/json/__init__.py
```diff
@@ -0,0 +1 @@
+# pybuild needs to add this directory even if it doesn't have .py files
```
m-roberts, line 1:
```suggestion
# This file needs to be here - Debian package building (pybuild) will only include directories that have at least one `.py` file.
```

pitop/robotics/json/alex.json
```diff
@@ -0,0 +1,43 @@
+{
+    "classname": "Pitop",
```
m-roberts, line 2: Configurations should have a schema version associated with them, in case of breaking changes in the future.

pitop/robotics/pan_tilt_controller.py
```diff
@@ -1,12 +1,143 @@
...
+    Recreatable,
+)
+from pitop.pma import ServoMotor
 
-class PanTiltController:
 
-    def __init__(self, servo_pan_port="S0", servo_tilt_port="S3"):
+class PanTiltController(Stateful, Recreatable):
+    CALIBRATION_FILE_DIR = ".config/pi-top/sdk"
+    CALIBRATION_FILE_NAME = "alex.conf"
```
m-roberts, line 19: This is technically part of configuration. This should be reviewed alongside how we are intending to handle the JSON files, to ensure that we have a solid plan for file management moving forwards.

pitop/robotics/pan_tilt_controller.py
```diff
@@ -1,12 +1,143 @@
...
+    def pan_servo(self):
+        return self._pan_servo
+
+    @property
+    @is_initialized
+    def tilt_servo(self):
+        return self._tilt_servo
+
+    @is_initialized
+    def calibrate(self, save=True, reset=False):
```
m-roberts, line 52: Whilst this does what it is supposed to, it's not very intuitive and lacks sophistication.

I wonder if we can/should create some sort of "pan-tilt calibrator", or similar that uses the miniscreen and buttons to simplify the process. It could be a separate program that is run - that is, part of the CLI - or something that can be called directly from Python. It would be great if a pi-top object could have its various subsystems calibrated, such as:
* iteratively tweaking a camera's resolution to optimise performance
* automatically determine tilt zero point by using camera as a sort of 'spirit level'
etc.

Just thinking out loud here, but I would love to see this improved.

examples/system/pitop_load_configuration.py
m-roberts: I would ensure that we have stock JSON files for different configurations saved somewhere, so that the examples can be run and work out-of-the-box. These can simply duplicate the `dict`s that are used internally.

pitop/pma/light_sensor.py
jcapona: nope, this will use the one in `ADCBase`

pitop/pma/encoder_motor.py
jcapona: agreed!

pitop/robotics/pan_tilt_controller.py
jcapona: Just created https://github.com/pi-top/pi-top-Python-SDK/issues/267 with your suggestions so we can work on it later on

angusjfw (commented):
👍 This is really great functionality to have and it seems like it works well. I've gone through this with Liv and other than some naming things we think the example usage is great. 

We discussed the relations and responsibilities of the mixins quite a bit as there are some non-obvious interactions there around the children list, method overrides and use of top level attributes and the name properties. We think it could be better if Stateful and Recreatable worked independently of Componentable. For example, we were thinking whether they could both keep their own arrays of child instances so as not to rely on the Componentable adding top level attributes. Maybe these mixins could be built on a common NodeTree mixin to standardise their interfaces with each other.

This is a great place to start though and we can work on this internal stuff incrementally as we use it more and it becomes clearer. 🚀

pitop/pma/led.py
```diff
@@ -1,19 +1,35 @@
...
+
+        Stateful.__init__(self)
+        Recreatable.__init__(self, {"port_name": port_name, "name": self.name})
+        gpiozero_LED.__init__(self, get_pin_for_port(self._pma_port), pin_factory=NativeFactory())
 
-        super(LED, self).__init__(get_pin_for_port(self._pma_port))
+    @property
+    def own_state(self):
+        return {
+            "is_lit": self.is_lit,
```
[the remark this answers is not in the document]
duwudi: Haha

examples/system/pitop_load_configuration.py
duwudi: Agree

pitop/camera/camera.py
duwudi: Yeah, I'm not a fan of it either, can we not just call it `state`? Or `self_state`

pitop/robotics/json/alex.json
```diff
@@ -0,0 +1,44 @@
+{
+    "version": "0.17.0",
+    "classname": "Pitop",
+    "module": "pitop.system.pitop",
+    "components": {
+        "camera": {
+            "index": 0,
```
duwudi, line 7: I've noticed sometimes the camera will go onto another index and I've had to manually change it to get it working again. I'd recommend we change the USB camera implementation to check other indexes if zero fails so that it's more robust - of course, if somebody were to build a robot config with two cameras then the index of each would need to be specified on the fly (and they will probably change randomly between zero and one so putting this in a config might be troublesome anyway).

jcapona: that's right, the device index can change... I think we tried probing indexes once but it was just to slow (probing + waiting to fail if it was an incorrect index), but I guess we can do it again checking indexes from 0 to 3, or something like that.

pitop/core/mixins/stateful.py
```diff
@@ -0,0 +1,39 @@
...
+        return {}
+
+    def __child_state(self, child_name):
+        child = getattr(self, child_name)
+        if hasattr(child, "component_state"):
+            return child.component_state
+        return None
+
+    @property
+    def component_state(self):
```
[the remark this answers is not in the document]
jcapona: yeeeah, I wanted to use `state` but there's already a property called `state` in the `ServoMotor` class and I didn't want to  overwrite it; I guess we could... will check

pitop/robotics/configurations.py
jcapona:
```
from pitop.robotics.configs import alex_configuration
p = Pitop.from_dict(alex_configuration)

pitop/robotics/json/alex.json
jcapona: captured in https://github.com/pi-top/pi-top-Python-SDK/issues/268

m-roberts (approved):

PR #262, opened 2021-03-05
Ensure webserver video endpoint sends each frame only once

angusjfw:
I saw discussion on the webserver pr #256 that the video was not working well. This is at least partly due to the generator function sending the frames running a non-blocked loop sending as often as possible, not just once for each frame. Sending frames repeatedly will waste time and resources that can be used delivery each frame promptly. I tested this briefly and it seems to have made the latency pretty low and consistent (when opening the webpage on my PC over wifi - VNC introduces quite a bit of latency)

### DIFF
--- a/pitop/webserver/run.py
+++ b/pitop/webserver/run.py
@@ -26,21 +26,22 @@ def command(ws):
     print('Command socket disconnected')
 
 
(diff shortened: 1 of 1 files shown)


jcapona (approved):
nice!

PR #263, opened 2021-03-08
Coloured Ball Detect and Pincers (Bobbie)

duwudi:
### TODO
* Resolve conversation on #328
* Resolve conversation here

#### Core changes
* Adds ball detection algorithm
* Adds Bobbie robotics configuration (I don't expect this to be used that much in the Further challenges)
* Adds pincer controller (no example?)

#### Example code
* Detect coloured balls

#### Tests
* Test ball detector (needs to reduce mocks)

#### Refactor
* Line detect reuses algorithms shared by ball detection
* `processing/utils` --> `processing/core`
* Renamed 'colour' to 'color' across the codebase
* DriveController

#### Peripheral changes
* USB camera rotation takes -270 to +270
* OLED accepts OpenCV images
* Updated wheel separation value in drive controller to be more accurate when performing `robot.drive.rotate`, exposed by property

#### Package changes
* Added imutils dependency

#### Misc changes
* gitignore

### Issues
* Colour filtering is quite slow (3 colours: 15 FPS. 1 colour: 20-22 FPS)

Likely due to `cv2.erode` and `cv2.dilate`, but these are necessary for good detections. I tried with all types of kernels but the default one is fastest. Gaussian blurring is also quite slow and this was changed to a simple `cv2.blur` (cannot remove blurring completely, unfortunately).

### Future work
* Servo calibration for pincer controller
* Update line follower: `line_detect` should be changed to a `LineDetector` class
* Improve FPS problem

We could possibly address the FPS issue by using `dlib`'s object tracker, with normal ball detect once every X frames/secs to prevent `dlib` indefinitely tracking false positives.

### Implementation Details
Ball detection works by finding the area and "roundness" of the shape (using `cv2.matchShapes`) and chooses the most likely ball from this data. It also completely rejects found contours that don't meet minimum criteria for a ball object. This was the simplest method that I could come up with turning to ML.


pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,231 @@
...
+
+            cv2.line(resized_frame, detection_points[colour][i - 1], detection_points[colour][i],
+                     draw_colour[colour], thickness)
+
+    if image_return_format.lower() != 'opencv':
+        robot_view = ImageFunctions.convert(resized_frame, format="PIL")
+    else:
+        robot_view = resized_frame
+
+    if len(colours) == 1:
```
duwudi, line 176: @jcapona when you get a minute, let's discuss this ball_detect return structure. It's a bit odd but it gave me what I wanted for the Further challenge - it almost definitely needs a refactor. Potentially a Ball class with `@dataclass` decorate that includes red, green and blue balls. The only issue is if they only pass one colour to find, it should be accessible just with e.g `ball.radius` instead of `ball.red.radius`

pitop/camera/camera_calibration/camera_calibration.py
```diff
@@ -0,0 +1,123 @@
+import numpy as np
```
m-roberts, line 1: is this something that is generically useful for multiple applications? i am wondering if we can have this handled by the CLI - `pi-top camera calibrate` type thing?

pitop/protoplus/adc.py
```diff
@@ -55,8 +55,8 @@ def poll(self, delay=0.5):
 
             results = self.read_all()
 
-            print("| " + str(results[0]) + "\t| " + str(results[1]) + "\t| " + str(results[2]) +
-                  "\t| " + str(results[3]) + "\t| " + str(results[4]) + "\t| " + str(results[5]) + "\t|")
+            print("| " + str(results[0]) + "\t| " + str(results[1]) + "\t| " + str(results[2])
```
m-roberts, line 58: Any chance you can revert all the changes that aren't related to your code?

m-roberts:
Would love some more write-up of what the thinking is with stuff here. I find that there are way too many comments in the code which would be better off in the description of this PR.

It would also help readability if anything that isn't specifically a code change for this PR was moved to a separate branch. Not a big issue on smaller PRs, but it makes navigating larger ones like this just that bit harder.

pitop/camera/camera_calibration/camera_calibration.py
duwudi: That's a good point actually yeah, let's do that!

pitop/protoplus/adc.py
duwudi: Yeah sorry, was trying to get Flake8 checks working and made these changes - I'll revert

pitop/camera/camera_calibration/camera_calibration.py
m-roberts: Was just thinking that this is a standalone application, same as the IMU calibration!

m-roberts: https://github.com/pi-top/pi-top-Python-SDK/issues/267

m-roberts:
Currently failing to build the docs due to:
```

  File "/usr/lib/python3/dist-packages/sphinx/cmd/build.py", line 304, in build_main
    app.build(args.force_all, filenames)
  File "/usr/lib/python3/dist-packages/sphinx/application.py", line 341, in build
    self.builder.build_update()
  File "/usr/lib/python3/dist-packages/sphinx/builders/__init__.py", line 347, in build_update
    len(to_build))
  File "/usr/lib/python3/dist-packages/sphinx/builders/__init__.py", line 360, in build
    updated_docnames = set(self.read())
  File "/usr/lib/python3.7/contextlib.py", line 119, in __exit__
    next(self.gen)
  File "/usr/lib/python3/dist-packages/sphinx/util/logging.py", line 261, in pending_warnings
    memhandler.flushTo(logger)
  File "/usr/lib/python3/dist-packages/sphinx/util/logging.py", line 224, in flushTo
    logger.handle(record)
  File "/usr/lib/python3.7/logging/__init__.py", line 1529, in handle
    self.callHandlers(record)
  File "/usr/lib/python3.7/logging/__init__.py", line 1591, in callHandlers
    hdlr.handle(record)
  File "/usr/lib/python3.7/logging/__init__.py", line 901, in handle
    rv = self.filter(record)
  File "/usr/lib/python3.7/logging/__init__.py", line 762, in filter
    result = f.filter(record)
  File "/usr/lib/python3/dist-packages/sphinx/util/logging.py", line 412, in filter
    raise SphinxWarning(message)
sphinx.errors.SphinxWarning: autodoc: failed to import class 'Pitop' from module 'pitop'; the following exception was raised:
No module named 'pitop.camera.camera_calibration'

Warning, treated as error:
autodoc: failed to import class 'Pitop' from module 'pitop'; the following exception was raised:
No module named 'pitop.camera.camera_calibration'
```

m-roberts:
If we are going to put the camera calibration stuff into another PR, then we should have an Issue to track that

jcapona (approved):

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,247 @@
...
+)
+from typing import Union
+from imutils import resize, grab_contours
+from imutils.video import FPS
+from pitop.core.data_stuctures import DotDict
+import atexit
+from pitop.processing.utils.vision_functions import import_opencv
+
+
+cv2 = import_opencv()
```
jcapona, line 16: please check https://github.com/pi-top/pi-top-Python-SDK/blob/master/pitop/camera/core/capture_actions/motion_detector.py#L19 to see how to use `import_opencv` with classes... 

If you run `import_opencv` at the top of a file that's imported whenever you run `import pitop`, it will throw an exception if `OpenCV` is not installed, that's why it was added as a class attribute, so it fails when you actually attempt to use `cv2`... you _could_
 use `cv2` as a global variable, initialized as `None`, and actually run `import_opencv` in the constructor.

On the examples is not that bad, since you know what to expect, and it should fail right away if a library is missing.

jcapona (changes requested):

m-roberts:
Before we merge this in, we should figure out what the line following code would look like, with the new dependencies that this PR brings #184

m-roberts:
> No support for servo calibration yet, this needs to be added once a branch with that on is merged into master

So 'servo calibration' needs to be added as an Issue and link to this PR.

examples/camera/camera_coloured_ball_detect.py
```diff
@@ -0,0 +1,44 @@
+from os import environ
+from pitopcommon.current_session_info import get_first_display
+from pitop.camera import Camera
+from pitop.processing.algorithms import BallDetector
+from signal import pause
+from pitop.processing.utils.vision_functions import import_opencv
+
+
+cv2 = import_opencv()
```
m-roberts, line 9: Probably don't want to encourage this for users who can just import it directly.

examples/camera/camera_coloured_ball_detect.py
```diff
@@ -0,0 +1,44 @@
...
+
+environ["DISPLAY"] = get_first_display()
```
m-roberts, line 11: Would be nice to review the common library soon - it was never intended to be used explicitly by users in this way. This is also only needed if run over SSH, and the solution to that is to run it with this environment variable set. I am not a huge fan of including it like this, although I do see how it is useful.

examples/camera/camera_coloured_ball_detect.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def process_frame(frame):
+    ball = ball_detector.detect(frame, colour=("red", "green", "blue"))
```
m-roberts, line 15:
```suggestion
    balls = ball_detector.detect(frame, colour=("red", "green", "blue"))
```

Especially if you're going to split it out after.

examples/camera/camera_coloured_ball_detect.py
```diff
@@ -0,0 +1,44 @@
...
+
+    print("--------- FOUND DATA -----------")
+    print(f"R: {red_ball.found} | G: {green_ball.found} | B: {blue_ball.found} \n")
+
+    print("--------- CENTER DATA ----------")
+    print(f"R: {red_ball.center} | G: {green_ball.center} | B: {blue_ball.center} \n")
+
+    print("--------- RADIUS DATA ----------")
+    print(f"R: {red_ball.radius} | G: {green_ball.radius} | B: {blue_ball.radius} \n")
+    print("\n")
```
m-roberts, line 34: This could be represented a little clearer, both in output and code.

pitop/camera/camera.py
```diff
@@ -32,10 +32,11 @@ class Camera(Stateful, Recreatable):
 
     def __init__(self,
                  index=None,
-                 resolution=None,
+                 resolution=(640, 480),
```
m-roberts, line 35: Why? This is an undocumented change

pitop/camera/core/cameras/usb_camera.py
```diff
@@ -41,14 +47,19 @@ def __del__(self):
...
             self.__camera.get_frame(),
             'raw',
             'RGB'
         )
 
+        if self._rotate_angle != 0:
+            pil_image = pil_image.rotate(angle=self._rotate_angle, expand=True)
+
+        return pil_image
+
```
m-roberts, line 62:
```suggestion
        return Image.frombytes(
            'RGB',
            (self.__camera.width, self.__camera.height),
            self.__camera.get_frame(),
            'raw',
            'RGB'
        ).rotate(angle=self._rotate_angle, expand=True)

```

m-roberts: Should be smart enough to just do nothing if angle is 0, so no need for storing to a variable

pitop/processing/algorithms/__init__.py
```diff
@@ -1 +1,2 @@
 from .line_detect import process_frame_for_line
+from .ball_detect import BallDetector
```
m-roberts, line 2: Should we make it clear that these are 'image processing' algorithms?

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,247 @@
...
+        self._process_image_width = process_image_width
+        self._input_format = input_format
+        self._output_format = output_format
+        self._frame_scaler = None
+        self._print_fps = print_fps
+        if self._print_fps:
+            self._fps = FPS().start()
+            atexit.register(self.__print_fps)
+
+    def detect(self, frame, colour: Union[str, tuple] = "red"):
```
m-roberts, line 81: This function is way too long for me. It should be broken up

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,247 @@
...
+                    # save to new max
+                    max_likelihood_index = likelihood_index
+                    # check if the ball is circular. Or if it's so big it's just close to the camera and occluded
+                    if match_value < ball_match_limits[colour] or radius > BALL_CLOSE_RADIUS:
+                        # check if it meets the minimum ball radius
+                        if radius > MIN_BALL_RADIUS:
+                            # ok, we're finally happy that this is an actual ball
+                            # store the values to be used as the most likely ball in frame
+                            ball_center = (int(x), int(y))
+                            ball_radius = int(radius)
```
m-roberts, line 245: Too many comments - would like to see more clarity provided by breaking up functions and providing good names for the discrete steps so that - at a glance - it is more obvious what is going on.

pitop/processing/algorithms/line_detect.py
```diff
@@ -1,13 +1,15 @@
-from numpy import arctan, array, pi
-
+from numpy import array
```
m-roberts, line 1: Nothing mentioned about line following in the PR description. Please amend the description to provide a complete description of the code changes proposed here!

pitop/processing/algorithms/line_detect.py
```diff
@@ -30,11 +32,11 @@ def calculate_blue_limits():
     return lower_blue, upper_blue
 
 
-def process_frame_for_line(frame, image_format="PIL", scale_factor=0.5):
+def process_frame_for_line(frame, image_format="PIL", process_image_width=320):
```
m-roberts, line 35: Why?

pitop/processing/utils/vision_functions.py
```diff
@@ -66,9 +66,34 @@ def find_largest_rectangle(rectangles):
...
+    if center is None:
+        return None
+    # convert so (0, 0) is at the middle bottom of the frame
+    center_x = center[0] - int(frame.shape[1] / 2)
+    center_y = int(frame.shape[0] / 2) - center[1]
+
+    return center_x, center_y
+
+
+def get_control_angle(center, frame):
```
m-roberts, line 86: Totally not obvious what this 'vision function' is doing, and why anyone would use it, or not - aside from the actual place where it's used...

pitop/robotics/drive_controller.py
```diff
@@ -28,16 +28,19 @@ def __init__(self, left_motor_port="M3", right_motor_port="M0", name="drive"):
         self.right_motor_port = right_motor_port
         self.left_motor_port = left_motor_port
 
-        # TODO: increase accuracy of wheel_base and wheel_diameter with empirical testing
-        self._wheel_separation = 0.1725
+        self._wheel_separation = 0.1675
```
m-roberts, line 31: Undocumented

setup.py
```diff
@@ -132,6 +132,11 @@
     "flask-sockets",
     "gevent",
     "gevent-websocket",
+
+    #############################
+    # Advanced image processing #
+    #############################
+    "imutils",
```
m-roberts, line 139: Is this a core dependency? How big is it?

tests/test_drive_controller.py
```diff
@@ -120,10 +120,10 @@ def test_motor_rpm_calculations_based_on_speeds(self):
 
         test_values = [
             [0, 0, 0, 0, 0],
-            [95.90, 114.00, 1, 1, 0],
-            [104.58, 114.00, 1, 1, 1],
-            [70.70, 84.10, 0.3, 0.3, 0],
-            [103.62, 114.0, 0.3, 0.3, 0.8],
+            [96.39, 114.00, 1, 1, 0],
```
m-roberts, line 123: To repeat myself - why?

jcapona: I updated these since these calculations were affected by the change in `DriveController._wheel_separation`

pitop/processing/algorithms/ball_detect.py
duwudi: ok got it

pitop/robotics/drive_controller.py
duwudi: By undocumented what do you mean? I added it to the description, should I be documenting it elsewhere too?

pitop/processing/algorithms/line_detect.py
duwudi: The scale factor is not useful since it will change performance depending on image passed in - best to scale to a known width so performance always gauranteed

pitop/processing/algorithms/line_detect.py
duwudi: There is one bullet point mentioning it, this is just cleaning up based on what has been developed since line_detect was made

pitop/processing/algorithms/ball_detect.py
duwudi: yeah agree, I mentioned I wasn't happy with this function to jorge but not sure how to improve it

pitop/processing/algorithms/__init__.py
duwudi: yeah we probably should, do you mean a separate directory or just adding comments in this __init__?

pitop/camera/camera.py
duwudi: So a user can just do `cam = Camera()` and not have to worry about resolution - especially useful for beginners. It's also good for us to have a standard resolution we develop algorithms for, so at least we know the main use-case

duwudi:
> > * No support for servo calibration yet, this needs to be added once a branch with that on is merged into master
> >   Servo calibration needs to be added as an Issue and link to this PR.
> 
> Before I feel like I can approve this PR, I would like to better understand how the code is expected to change in the future around this code. Are we going to switch to `dlib`? I think that this relates to #184 as well, as it would improve the code there. How likely is it that the code here is going to break/change?

I've thought about dlib object tracker a lot for ball detection and I don't think it's the right approach. It's too easy for the algorithm to get false detections and continue to track them - since dlib object tracker is essentially a black box we'd have to build workarounds like running our detection once every 5 frames or having a lower limit on dlib's confidence at which we run our detection again - all of which won't be that robust in a world where colour and "roundness" is our only detection criteria. dlib's object tracker works well for things that are very unique, such as a face or human body - I'm sure it can work here too, but I don't think it would be optimum. 

I think a better approach is a Kalman filter- it's simpler, faster and can be tuned better for noisy measurement rejection on the fly (rather than being in a "detect" state or a "dlib track" state, we'd just be updating our predictions based on new data and rejecting or using them). I've already started looking into this approach and it will work much better in other ways too, such as being able to track the ball off-frame for a short period of time (dead-reckoning) so the robot doesn't just lose the object if it turns incorrectly for a brief moment - this will be super useful for line tracking too.

I don't think any filtering will introduce breaking changes as it should just be optimization internally, as long as we are happy with:
- Return parameters
- Public function names
- The class name
Then I think it's fine

setup.py
duwudi: It's pretty useful, I find myself needing more of his stuff all the time. We could just copy and paste the stuff we need into our own lib (I did this before) but it's a nice lib that people use a lot for opencv so I think it's worth keeping. I believe it's around 50kb

examples/camera/camera_coloured_ball_detect.py
duwudi: Shall I remove it?

pitop/processing/algorithms/ball_detect.py
duwudi: See updates

pitop/processing/algorithms/ball_detect.py
duwudi: Refactored, let me know what you think

pitop/processing/utils/vision_functions.py
duwudi: Added description and tried to make it clearer

examples/camera/camera_coloured_ball_detect.py
m-roberts: i think that we should probably add a section to the docs that explains the extra steps to take when you are running code over SSH. that way we can keep the code minimal, as if it's being run from the device - which would mean removing it here

tests/test_ball_detector.py
```diff
@@ -0,0 +1,166 @@
...
+    "pitopcommon.logger",
+    "pitopcommon.singleton",
+    "pitopcommon.common_ids",
+    "pitopcommon.current_session_info",
+    "pitopcommon.ptdm",
+    "pitopcommon.firmware_device",
+    "pitopcommon.command_runner",
+    "pitopcommon.common_names"
+
+]
```
m-roberts, line 21: I don't think we need all of these for the ball detector. Let's remove the ones we don't need.

examples/camera/camera_coloured_ball_detect.py
m-roberts: This has been moved to a separate issue: https://github.com/pi-top/pi-top-Python-SDK/issues/325

pitop/camera/camera.py
m-roberts: Let's drop this from here and implement a better solution in this Issue: https://github.com/pi-top/pi-top-Python-SDK/issues/326

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,279 @@
...
+
+MIN_BALL_RADIUS = 5
+BALL_CLOSE_RADIUS = 50
+DETECTION_POINTS_BUFFER_LENGTH = 64
+
+
+class BallDetector:
+    def __init__(self,
+                 process_image_width: int = 320,
+                 input_format: str = "PIL",
```
m-roberts, line 56: Drop this

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,279 @@
...
+                 output_format: str = "PIL",
```
m-roberts, line 57: `format`

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,279 @@
...
+    ],
+    'green': [{
+        'lower': (60, 100, 100),
+        'upper': (90, 255, 255)
+    }],
+    'blue': [{
+        'lower': (100, 100, 100),
+        'upper': (130, 255, 255)
+    }]
+}
```
m-roberts, line 34: This could become more globally useful

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,275 @@
...
+        self._fps.stop()
+        print(f"[INFO] Elapsed time: {self._fps.elapsed():.2f}")
+        print(f"[INFO] Approx. FPS: {self._fps.fps():.2f}")
+
+    def __prepare_frames(self, frame):
+        frame = ImageFunctions.convert(frame, format='OpenCV')
+
+        if self._frame_scaler is None:
+            height, width = frame.shape[0:2]
+            self._frame_scaler = width / self._process_image_width
```
m-roberts, line 156: Out of scope of function - can we move it somewhere else?

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,275 @@
...
+            raise ValueError('Cannot pass more than three colours.')
+
+        return colours
+
+    def colour_filter(self,
+                      frame,
+                      colour: str = "red",
+                      return_binary_mask: bool = False,
+                      input_format: str = "PIL",
+                      output_format: str = "PIL"
```
m-roberts, line 198: `format`

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,275 @@
...
+            colours = colour
+        if len(colours) > 3:
+            raise ValueError('Cannot pass more than three colours.')
+
+        return colours
+
+    def colour_filter(self,
+                      frame,
+                      colour: str = "red",
+                      return_binary_mask: bool = False,
```
m-roberts, line 196: PIL uses `mode="1"` for 1-bit images - perhaps we can use that?

m-roberts: Concluded we will remove from the function, apply after.

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,275 @@
...
+
+        return mask
+
+    def __find_contours(self, frame, colour):
+        mask = self.colour_filter(frame, colour=colour, return_binary_mask=True, input_format="OpenCV",
+                                  output_format="OpenCV")
+
+        contours = self.cv2.findContours(mask.copy(), self.cv2.RETR_EXTERNAL, self.cv2.CHAIN_APPROX_SIMPLE)
+        contours = grab_contours(contours)  # fixes problems with OpenCV changing their protocol
+        return contours
```
m-roberts, line 235:
```suggestion
        return grab_contours(contours)  # fixes problems with OpenCV changing their protocol
```

examples/camera/camera_coloured_ball_detect.py
```diff
@@ -0,0 +1,30 @@
+from pitop.camera import Camera
+from pitop.processing.algorithms import BallDetector
+from signal import pause
+import cv2
+
+
+def process_frame(frame):
+    detected_balls = ball_detector.detect(frame, color=("red", "green", "blue"))
```
m-roberts, line 8:
```suggestion
    detected_balls = ball_detector.detect(frame, color=["red", "green", "blue"])
```

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,269 @@
...
+        return self.center is not None
+
+
+class BallDetector:
+    def __init__(self,
+                 process_image_width: int = 320,
+                 format: str = "OpenCV",
+                 print_fps: bool = False):
+        """
+        :param int process_image_width: image width to scale to for image processing
```
m-roberts, line 89: Not my most favourite variable name

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,269 @@
...
+        self._print_fps = getenv("PT_ENABLE_FPS", "0") == "1"
+        if self._print_fps:
+            self._fps = FPS().start()
+            atexit.register(self.__print_fps)
+
+    def detect(self, frame, color: Union[str, list] = "red"):
+        def parse_colors(color_arg):
+            colors = []
+            if type(color_arg) == str:
+                assert(color_arg in valid_colors)
```
m-roberts, line 108: Instead of asserting, we should probably check and raise an Exception

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,269 @@
...
+    def color_filter(self,
+                     frame,
+                     color: str = "red",
+                     return_binary_mask: bool = False,
+                     format: str = "PIL"
+                     ):
+
+        frame = ImageFunctions.convert(frame, format="OpenCV")
+
+        if color.lower() not in ("red", "green", "blue"):
```
m-roberts, line 171: This is not _reaaaally_ what a tuple is for - (r, g, b), (x, y) and the like. This would be better as a list. Also, we check for this with `valid_colors` at the top - we might want to create a function to do this check as we have 2 places that use it.

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,269 @@
...
+            hsv_lower = color_range["lower"]
+            hsv_upper = color_range["upper"]
+            mask = self.cv2.inRange(hsv, hsv_lower, hsv_upper)
+            masks.append(mask)
+        mask = sum(masks)
+
+        mask = self.cv2.erode(mask, None, iterations=1)
+        mask = self.cv2.dilate(mask, None, iterations=1)
+
+        if not return_binary_mask:
```
m-roberts, line 188: As we discussed, this logic needs to be moved out - the mask should be applied to the image explicitly.

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,269 @@
...
+class BallDetector:
+    def __init__(self,
+                 process_image_width: int = 320,
+                 format: str = "OpenCV",
+                 print_fps: bool = False):
+        """
+        :param int process_image_width: image width to scale to for image processing
+        :param str format: output image format
+        """
+        self.cv2 = import_opencv()
```
m-roberts, line 92: Why does this need to be here? If this algorithm is being imported then we know it is needed.
Simply putting `cv2 = import_opencv()` at the top would do it, right?

examples/camera/camera_coloured_ball_detect.py
```diff
@@ -5,18 +5,21 @@
...
-    red_ball = balls.red
-    print(f'Red ball center: {red_ball.center}')
+    red_ball = detected_balls.red
+    if red_ball.is_valid():
+        print(f'Red ball center: {red_ball.center}')
 
-    green_ball = balls.green
-    print(f'Green ball center: {green_ball.center}')
+    green_ball = detected_balls.green
+    if green_ball.is_valid():
```
duwudi, line 15: I prefer `red_ball.found` as it's more readable to a beginner when reading an if statement - thoughts?

pitop/processing/algorithms/ball_detect.py
duwudi: I was told to put it as a class member to prevent cv2 errors if someone tries to import another file that triggers one of the __init__ files to import this one as a side effect

pitop/processing/algorithms/ball_detect.py
```diff
@@ -60,133 +92,72 @@ def __init__(self,
...
-            if ball_center is not None:
-                ball_center, ball_radius = self.__reposition_to_original_frame_size(ball_center, ball_radius)
+    def detect(self, frame, color: Union[str, list] = "red"):
+        def parse_colors(color_arg):
+            colors = []
+            if type(color_arg) == str:
+                assert(color_arg in valid_colors)
+                colors = [color_arg]
+            elif type(color_arg) in (list, tuple):
+                assert(set(color_arg).issubset(valid_colors))
```
duwudi, line 90: Nice!

pitop/processing/algorithms/ball_detect.py
```diff
@@ -60,133 +92,72 @@ def __init__(self,
...
 
-        if self.format.lower() != 'opencv':
-            robot_view = ImageFunctions.convert(robot_view, format="PIL")
+        robot_view = frame.copy()
+        ball_data = DotDict({})
+        for ball_color, ball_object in self.balls.items():
+            ball_data[ball_color] = ball_object
+            if ball_object.is_valid():
+                self.__draw_ball_position(robot_view, ball_object)
+                self.__draw_ball_contrail(robot_view, ball_object)
```
duwudi, line 114: This should be drawn regardless of if the ball object is valid, it will "fizzle" out anyway since "None" will be loaded into the deque of many ball detections are not found - this is the behaviour we want

pitop/processing/algorithms/ball_detect.py
```diff
@@ -234,26 +205,54 @@ def __find_contours(self, frame, color):
             )
         )
 
-    def __find_most_likely_ball(self, color, resized_frame):
-        ball_center = None
-        ball_radius = None
+    def __find_most_likely_ball(self, ball, color, frame):
+        if ball is None:
```
duwudi, line 208: Why would this ever be None?

tests/test_ball_detector.py
duwudi: @m-roberts can you take a stab at this since you know the architecture better? I'd just be blindly removing and pushing to test which might take a while

pitop/processing/algorithms/ball_detect.py
```diff
@@ -0,0 +1,273 @@
...
+    def __find_most_likely_ball(self, ball, color, frame):
+        resized_frame = resize(frame, width=self._image_processing_width)
+        contours = self.__find_contours(resized_frame, color)
+        if len(contours) == 0:
+            ball.no_detection()
+            return ball
+
+        max_likelihood_index = 0
+
+        def __meets_minimum_ball_requirements(_ball, _match_radius, _match_value):
```
m-roberts, line 229: Why are these prefixed with underscore?

m-roberts: I guess to prevent shadow of outer variable..

jcapona (approved):

Issue #268, opened 2021-03-15
Camera: try different device indexes on instantiation if default fails


jcapona:
When instantiating a `Camera` object the default device index used is 0, since that's the device index assigned to the first camera connected to the system. However, on certain circumstances, this index changes and can be > 0, even though there's only 1 camera connected.

To make the `Camera` class more robust, the constructor should probe different devices indexes if the default fails, so users don't have to manually specify it/change it.

We need to consider how this affects the configuration of our `Camera` component for recreating it.

PR #271, opened 2021-03-16
Loop over available devices to create camera object

jcapona:
Closes https://github.com/pi-top/pi-top-Python-SDK/issues/268

- If no specific camera device index is provided, loop over available devices trying to create a camera object.
- The index is no longer stored in the camera state, unless it's explicitly set by the user.

### DIFF
--- a/pitop/camera/core/cameras/usb_camera.py
+++ b/pitop/camera/core/cameras/usb_camera.py
@@ -1,22 +1,34 @@
 from PIL import Image
+from os import listdir, system
 
 from PyV4L2Camera.camera import Camera as V4L2Camera
 from PyV4L2Camera.exceptions import CameraError as V4L2CameraError
 
 
 class UsbCamera:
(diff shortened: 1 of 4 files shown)


pitop/camera/core/cameras/usb_camera.py
```diff
@@ -31,3 +43,25 @@ def get_frame(self):
...
+
+    @staticmethod
+    def list_device_indexes():
+        indexes = []
+        device_names = [name for name in listdir("/dev") if "video" in name]
+
+        # Not all devices actually provide video
+        # eg: video1 can be a metadata device for video0
+        for device in device_names:
+            cmd = f"v4l2-ctl --list-formats --device /dev/{device} | grep -qE '[[0-9]]'"
```
jcapona, line 55: TODO: add package that installs `v4l2-ctl` as a dependency

m-roberts (approved):

PR #293, opened 2021-03-29
Face detection/emotion tracking robot (Prax)

duwudi:
### TODO: spec defined
* ~~Write parser for `dlib` models that:~~
    * ~~checks in `~/.config/pi-top/sdk/dlib_models/`, and loads if found~~
    * ~~automatically attempts to download the model from https://github.com/davisking/dlib-models if not found locally (with print progress)~~ Extraction of bz2 currently doesn't have a progress bar but only takes around 10s
    * ~~throws exception if not able to retrieve~~ ** if the download fails, an exception is raised. No error checking on extraction because it's unlikely to fail - corrupt downloads are stored in a tmp file so won't be inadvertently used by extraction method

### TODO: spec TBD
* ~~Rewrite PR description explaining what is happening at a high level (making reference to line numbers and classes directly where useful), and explain as much of the motivation and architectural decisions as possible in order to make this PR reviewable. try not to simply provide a summary of what can be deduced at-a-glance from the code~~
* Propose change(s) that would allow people to write classifiers, detectors or trackers for specific objects and/or use an existing model

## **FaceDetector**
- Uses dlib frontal face detector and the dlib facial landmarks predictor. The latter requires the use of `shape_predictor_68_face_landmarks.dat` (100MB) from https://github.com/davisking/dlib-models
- Added dlib correlation tracker to track detected face - increases FPS from 6.33 to 11.47 (compared to detecting face every frame). Has the added benefit of being able to track the face at longer distances once an initial detection is found.
- The face tracking will continue until the peak-to-side-lobe ratio reported by dlib's correlation object tracker is less than 7.0 (https://www.cs.colostate.edu/~draper/papers/bolme_cvpr10.pdf). If it is less than 7.0, the class will revert back to full face detection to attempt to find the face again before returning to the user.
- the __call__ function takes a frame and uses either the full face detection method or object tracking depending on if a face was detected in the previous frame.
- Returns a Face object that can be used directly by the user or passed into other classes/functions that do additional processing (emotions, drowsiness, image overlays, age detection etc). Architecturally, we could convert these additional functions to mixins to the face detector class, but this would further abstract was is happening and make the return data structure more complex. Also, I'd rather the user do this procedurally in their code so they understand data flow from one thing to another and can understand conceptually what each "thing" is doing.

## **EmotionClassifier**
- Takes the data returned by FaceDetector to do additional processing that detects emotions. Currently, it takes one argument `face` and this `Face` object includes the original image frame used for face detection. We could change this so that the user passes in both a frame and a `Face` object but since it always needs to be the original frame and it's fully dependant on the FaceDetector I don't think this is necessary (although I admit that storing the original detection frame in a Face object does seem a little odd)
- Uses a custom trained SVC to detect between 6 emotions with optional mean filtering that can be switched off in the case of using static images as opposed to a video stream.
- Requires onnxruntime to be installed
- See docstrings in EmotionDetector class for more information on how the classification process works.

## **TiltRollHeadController**
- Used to control the servos in the Prax configuration 
- Includes a PID to imitate a user's face angle with Prax's face

## **PanTiltObjectTracker**
- Added class inside PanTiltController for tracking an object with a camera mounted on the pan-tilt servo mechanism - very useful when paired with the face detector

## **Added dependencies**
- **shape_predictor_68_face_landmarks.dat** - used for face angle and emotion classification
- **shape_predictor_5_face_landmarks.dat** - can also be used for face angle. In the case where emotion classification is not needed, the performance will be slightly better for face detection when using this 5 landmark version instead of the 68 landmark version  - an initialisation parameter has been added to allow users to do this so we should also make this file downloadable. 
- **dlib** (which requires liblapack-dev, libatlas-base-dev)
- **onnxruntime** - emotion classifier model format
- **fonts-symbola** - used for emoji's displayed on miniscreen

## **Other**
- Camera horizontal and vertical flip - Prax's camera is facing the user to a flip rather than a rotation is desired (so your face moves as though you're looking at a mirror)
- DotDict data structure moved to pi-top core
- Camera calibration parameter loader  - the ability to carry out camera calibration will move to CLI in future, but I intend to use this preconfigured calibration file soon so we can keep it here for now. It's purely internal so can be moved at any time without introducing breaking changes.
- Example recipe that shows the facial landmarks and emotion on the miniscreen
2021-04-08 16_55_45-out1 gif ‎- Photos


pitop/robotics/tilt_roll_head_controller.py
```diff
@@ -0,0 +1,167 @@
...
+        current_angle = self.roll_servo.current_angle
+        state = current_angle - angle
+        if abs(state) < 1.0:
+            self.roll_servo.sweep(speed=0)
+        else:
+            servo_speed = self._head_roll_pid(state)
+            self.roll_servo.sweep(speed=servo_speed)
+
+    @is_initialized
+    def calibrate(self, save=True, reset=False):
```
duwudi, line 75: need to delete calibration from here, too much duplication with pantiltcontroller

pitop/processing/algorithms/faces/face_detector_haarcascade.py
```diff
@@ -0,0 +1,121 @@
...
+        h_new = min(self._frame_resolution[1] - y_new, h_new)
+
+        return x_new, y_new, w_new, h_new
+
+    @staticmethod
+    def __draw_on_frame(frame, center, rectangle):
+        x, y, w, h = rectangle
+        cv2.rectangle(frame, (x, y), (x + w, y + h), (255, 0, 0), 2)
+        cv2.drawMarker(frame, center, (100, 60, 240), markerType=cv2.MARKER_CROSS, markerSize=20,
+                       thickness=4, line_type=cv2.FILLED)
```
duwudi, line 121: May delete this, dlib one is much better

pitop/processing/algorithms/faces/face_detector.py
```diff
@@ -0,0 +1,114 @@
...
+from pitop.core import ImageFunctions
+from imutils import face_utils
+
+# directory where calibration output pickle file is located
+classifier_dir = 'predictors'
+script_dir = os.path.dirname(os.path.realpath(__file__))
+abs_file_path = os.path.join(script_dir, classifier_dir)
+
+# Filename for predictor
+predictor_file_name = "shape_predictor_68_face_landmarks.dat"
```
duwudi, line 18: This is ~100MB, but is necessary for the emotion/expression classification which will be done soon

pitop/processing/algorithms/faces/face_detector_haarcascade.py
```diff
@@ -0,0 +1,121 @@
+import os
+import cv2
```
jcapona, line 2: use `import_opencv` from `pitop.processing.utils.vision_functions` since opencv might not be installed... check `pitop/processing/algorithms/line_detect.py` to see how it's used

pitop/processing/algorithms/faces/face_detector_haarcascade.py
```diff
@@ -0,0 +1,121 @@
...
+# directory where calibration output pickle file is located
+classifier_dir = 'classifiers'
+script_dir = os.path.dirname(os.path.realpath(__file__))
+abs_file_path = os.path.join(script_dir, classifier_dir)
+
+# Filename for classifier
+cascade_model_classifier = 'haarcascade_frontalface_default.xml'
+
+
+class DotDict(dict):
```
jcapona, line 19: lets put this into a "utils" file maybe in `pitop.core` since it's used in 2 places now

pitop/processing/utils/vision_functions.py
```diff
@@ -1,3 +1,12 @@
+
+try:
+    import cv2
+except (ImportError, ModuleNotFoundError):
+    raise ModuleNotFoundError(
+        "OpenCV Python library is not installed. You can install it by running "
+        "'sudo apt install python3-opencv libatlas-base-dev'.") from None
```
jcapona, line 7: lets not do this

jcapona (changes requested):

pitop/processing/algorithms/faces/face_detector.py
duwudi: can we download it upon request? In case someone tried to do the challenge. Also, I'm wondering if we can compress it and decompress upon installation

pitop/processing/algorithms/faces/face_detector_haarcascade.py
duwudi: yeah agree

pitop/processing/utils/vision_functions.py
```diff
@@ -67,8 +72,56 @@ def find_largest_rectangle(rectangles):
...
 
 def scale_frame(frame, scale):
-    cv2 = import_opencv()
     scaled_width = int(frame.shape[1] * scale)
     scaled_height = int(frame.shape[0] * scale)
     dim = (scaled_width, scaled_height)
     return cv2.resize(frame, dim, interpolation=cv2.INTER_AREA)
+
+
+def resize(image, width=None, height=None, inter=cv2.INTER_AREA):
```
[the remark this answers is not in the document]
duwudi: @jcapona do we really want to import opencv every time this function is called? I don't mind doing this for single-call function during initialization but it seems like it will bring a massive overhead when doing it 30-60 times per second in a frame callback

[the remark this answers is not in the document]
duwudi: I read somewhere that having the cv2.INTER_AREA in the parameter itself improves efficiency, I'll dig it up - might not be necessary for us but interesting nonetheless

pitop/processing/algorithms/faces/face_detector.py
m-roberts: How is this file provided?

pitop/processing/utils/vision_functions.py
m-roberts: Let's do what then?

pitop/processing/utils/vision_functions.py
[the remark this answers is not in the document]
m-roberts:
```
pi@pi-top:~ $ time python3 -c "import cv2"

real	0m2.902s
user	0m2.706s
sys	0m0.181s
pi@pi-top:~ $ time python3 -c "import cv2; import cv2"

real	0m2.926s
user	0m2.741s
sys	0m0.180s
pi@pi-top:~ $
```

Seems pretty low cost to perform additional imports.

debian/control
```diff
@@ -27,6 +27,10 @@ Depends:
  ${misc:Depends},
  python3-opencv,
  python3-pitop (= ${binary:Version}),
+ libopenblas-dev,
+# Required for face detection & emotion mapping
```
m-roberts, line 31: Which is not a dependency of the core SDK. It's part of `-full`

debian/py3dist-overrides
```diff
@@ -10,3 +10,6 @@ python_sonic python3-sonic
 pyv4l2camera python3-pyv4l2camera
 scipy python3-scipy
 simple_pid python3-simple-pid
+dlib python3-dlib
+scikit-learn python3-sklearn
```
m-roberts, line 14: What is this for?

pitop/processing/utils/vision_functions.py
m-roberts: Agree - revert this and leave the imports

setup.py
```diff
@@ -123,6 +123,13 @@
     # Algorithms #
     ##############
     "simple_pid",
+
+    ###########################################
+    # Face detection & emotion classification #
```
m-roberts, line 128: "Advanced image processing"

m-roberts:
Overall, I find this to be hard to read and unclear in its scope.

I find this face tracking implementation to be sufficient:
https://www.guidodiepen.nl/2017/02/detecting-and-tracking-a-face-with-python-and-opencv/

Also, the paradigm implemented (with its use of `dlib`'s correlation tracker) can be easily used for any form of tracking (such as line following).

I would like to see either a stripping back or a clear explanation of what this PR is doing that justifies the extra complexity.

setup.py
```diff
@@ -123,6 +123,13 @@
...
+    ###########################################
```
m-roberts, line 129:
```suggestion
    ###########################################
    # Advanced image processing #
    ###########################################
```

debian/py3dist-overrides
duwudi: It's to run an SVC for emotion classification - I figured we'd want it on the OS anyway since we plan on doing a lot more AI related content

duwudi: I think jorge added this after we discussed the emotion classification code, which isn't actually a part of this PR. I'll clean up the two branches and make the scope clearer

debian/control
jcapona: not sure what you mean @m-roberts

examples/recipes/robot_display_emotions.py
```diff
@@ -0,0 +1,147 @@
...
+    "Surprise": u"\U0001F62E",
+    "Neutral": u"\U0001F610",
+    "Sad": u"\U0001F641",
+    "Disgust": u"\U0001F62C",
+    "Looking": u"\U0001F440"
+}
+
+
+text_font = ImageFont.truetype("VeraMono.ttf", size=12)
+emoji_font = ImageFont.truetype("Symbola.ttf", size=50, encoding='unic')
```
duwudi, line 33: @jcapona how can we add this font to the OS? It's a good one for displaying emoji's on the miniscreen

duwudi: Yeah I think when I get around to writing a Further challenge for Prax I will definitely make a draw emoji function so let's just add it in for now

pitop/robotics/drive_controller.py
```diff
@@ -7,7 +7,6 @@
 
 from simple_pid import PID
 
-from pitop.core.exceptions import UninitializedComponent
```
duwudi, line 10: @jcapona any reason this hasn't been done to master branch yet? Or is it just something we're going to merge in with this PR? I'm not entirely sure of the implications

debian/control
m-roberts: Face detection and emotion mapping are not core dependencies of the SDK. They are dependencies of OpenCV, so should move to `-full`.

markpitop:
@m-roberts, @angusjfw 
The question @ryan asked above, "do we use dlib", is pretty important.  It's a popular, maintained, respected lib from what I can see, and there are lots of docs talking about face detection.  I've found some repos with other models (dogs, cars) that we can use.  I'm not sure of the following:
1. How hard it is to create new models or if we'd even want to.
2. If the detectors in the latest opencv are good enough (they're supposed to be faster). I.e. Could we simply use opencv if we upgrade that?
3. What sort of space / load does dlib represents?
4. If there's something else we should be considering?
The problem I see is that if challenges are written against this in Further, then it because a burden to maintain and hard to change. It no longer matters if this is in labs, it's going to get baked in. 
I think we need to resolve questions relating to this asap. What's the best approach?

examples/camera/emotion_test.py
```diff
@@ -0,0 +1,23 @@
+from pitop.processing.algorithms.faces import FaceDetector, EmotionDetector
+from pitop.processing.core.vision_functions import import_opencv
+
+
+cv2 = import_opencv()
+
+face_detector = FaceDetector()
```
markpitop, line 7: Spitballing here. I think we'd aspire to:
face_detector = Detector(face_model)
dog_detector = Detector(dog_model)
it's more flexible, and allows us to explore models more easily with students.
I haven't reviewed FaceDetector yet, so may change / delete this comment later :)
Anyone else have thoughts?

m-roberts:
This PR still has `joblib` model and test dependency included. If we are moving to ONNX, we don't want this.

examples/camera/emotion_test.py
duwudi: something like `face_detector = ObjectDetector(detectors.FACE)` could work, but all it would be doing is giving them the `FaceDetector` class (or `BallDetector`). The issue is that each one will have a very different implementation, it's not just a case of passing a `model` to it

markpitop: let's have a chat - I need to understand that better. Do we have BallDetector already?

angusjfw: I think the important thing is not to share implementations but provide a consistent interface to our image processing & feature detector functionality, so that they could be composed in flexible ways.

For example feature detectors should take an image and return a standardised result including a processed image ('robot view') and an array of features, with their positions in the image and confidence values. Those features can then include the non-standard information such as an instance of a 'Face' class with features etc. This should be applicable to our ball, line and face detectors and will help people to use them outside the provided example code because they follow a simple pattern.

Similarly I don't think it's correct for our ball and face detectors to be classes whilst line detector is a function. None of them require tracking previous frame or other state that I can see, so they don't need to be classes, though perhaps it's worth using a class for the one-time configuration and fps tracking features (which certainly could be abstracted into a Detector super class). Motion detection is an example where tracking state from previous frames is required.

This organisation helps to gain insight for ourselves and users for example that emotion detection doesn't fit this pattern the same way as these other detectors. It doesn't use the image as part of it's processing, only for annotation, so perhaps we consider it to be a 'classifier' instead, possibly embedded in the face detector as it always requires that.

duwudi: Both face detector and emotion detector require state, line detector should be changed to a class too as then we can use a Kalman filter to track the line’s position even when it goes slightly off frame - this will also help to minimise false detection too and generally make it more robust.

I agree with the other comments though, standardised return data structure would be nice - I’ve seen this done if other libs that we can take inspiration from. The reason I didn’t embed emotion detector in face detector is because I’d like to have other things like a “drowsiness detector” that also takes a face as input (also needs to be a class since they’ll be filtering there too). I guess there’s multiple ways to do this though so would love to discuss it more

examples/recipes/robot_display_emotions.py
m-roberts: `Symbola_hint.ttf` works though. the package only installs 1 font file:
```
/usr/share/fonts/truetype/ancient-scripts/Symbola_hint.ttf
```

pitop/camera/camera.py
```diff
@@ -32,10 +32,12 @@ class Camera(Stateful, Recreatable):
 
     def __init__(self,
                  index=None,
-                 resolution=None,
+                 resolution=(640, 480),
                  camera_type=CameraTypes.USB_CAMERA,
                  path_to_images="",
                  format='PIL',
+                 flip_top_bottom: bool = False,
+                 flip_left_right: bool = False,
```
m-roberts, line 40: No.

duwudi: https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.transpose - it's needed to flip a camera frame from top to bottom or left to right. Prax's camera is mounted in a flipped orientation and because it's facing the user it needs to be flipped (mirrored) and not just rotated

debian/python3-pitop-full.install
```diff
@@ -0,0 +1,2 @@
+pitop/processing/algorithms/faces/predictors	/usr/lib/python3/dist-packages/pitop/processing/algorithms/faces/predictors
+pitop/processing/algorithms/faces/models      /usr/lib/python3/dist-packages/pitop/processing/algorithms/faces/models
```
jcapona, line 2: these should be removed once we tackle the large-files issue

pitop/processing/algorithms/faces/emotion_classifier.py
```diff
@@ -0,0 +1,202 @@
...
+from pitop.processing.core.vision_functions import (
+    import_opencv,
+    tuple_for_color_by_name,
+)
+
+
+cv2 = import_opencv()
+
+
+class Emotion:
```
jcapona, line 15: lets move this class into a separate file

pitop/processing/algorithms/faces/face_detector.py
```diff
@@ -0,0 +1,326 @@
...
+
+cv2 = import_opencv()
+dlib = import_dlib()
+
+predictor_dir = 'predictors'
+script_dir = path.dirname(path.realpath(__file__))
+abs_file_path = path.join(script_dir, predictor_dir)
+
+
+class Face:
```
jcapona, line 29: lets move this class into a separate file

pitop/robotics/drive_controller.py
jcapona: sorry... yeah, this should be merged with this PR... if you want we can separate it into another one to avoid confussion

pitop/robotics/pan_tilt_controller.py
```diff
@@ -1,56 +1,104 @@
...
     Recreatable,
 )
+from pitop.robotics.two_servo_assembly_calibrator import TwoServoAssemblyCalibrator
 from pitop.pma import ServoMotor
+from pitop.pma.servo_controller import ServoHardwareSpecs
+from simple_pid import PID
+from time import time
+
+
+class PanTiltObjectTracker:
```
jcapona, line 12: lets move this class into a separate file

jcapona (changes requested):
Mostly moving classes to separate files, I think it's ok if they're small, but in this case they're quite long... 

Now, regarding the large files, I think we should just download them from the original GitHub repository when used, and store them somewhere so they can be reused later on (I think we discussed that with @m-roberts ) ... Of course, this would depend on the user having an internet connection.

We could even have a pi-top repository with all the other large files/models/calibrations, I think it might be helpful for someone out there...

pitop/camera/camera.py
duwudi: added to description

debian/python3-pitop-full.install
duwudi: Do I just delete these lines or the whole file? Also, I've kept the onnx model in the repo but it is 2.6MB - maybe we should create a separate repo for this and download it like the dlib ones?

duwudi: Also, would you mind explaining what these lines are even doing?

duwudi: I've added the onnx model here: https://github.com/pi-top/pi-top-SDK-models

pitop/processing/algorithms/faces/face_detector.py
```diff
@@ -0,0 +1,245 @@
...
+            current_area = w * h
+            if current_area > area:
+                area = current_area
+                largest_rectangle_dlib = rectangle_dlib
+                face_rectangle = (x, y, w, h)
+                face_center = (int(x + w / 2), int(y + h / 2))
+
+        if rectangles_dlib is None:
+            # this should never trigger. Discuss.
+            return None, None, None
```
jcapona, line 151: if this should never trigger, why not just raise an exception? something must be wrong somewhere, right?

duwudi: I've removed it completely. It's basically impossible for it to trigger unless the width and height of the returned rectangle is 0, which I think we can safely assume will never be the case

duwudi: Also FYI this was supposed to be checking `largest_rectangle_dlib`, not `rectangles_dlib` - that was a typo anyway and was even less likely since dlib face detector always returns a list

jcapona (approved):

PR #328, opened 2021-04-14
Bobbie changes

m-roberts:
This PR is to discuss changes since an earlier commit on Bobbie with a simpler diff.

### DIFF
--- a/examples/camera/camera_coloured_ball_detect.py
+++ b/examples/camera/camera_coloured_ball_detect.py
@@ -8,19 +8,19 @@ def process_frame(frame):
     detected_balls = ball_detector.detect(frame, color=["red", "green", "blue"])
 
     red_ball = detected_balls.red
-    if red_ball.is_valid():
+    if red_ball.found:
         print(f'Red ball center: {red_ball.center}')
         print(f'Red ball radius: {red_ball.radius}')
         print(f'Red ball angle: {red_ball.angle}\n')
 
     green_ball = detected_balls.green
-    if green_ball.is_valid():
+    if green_ball.found:
         print(f'Green ball center: {green_ball.center}')
         print(f'Green ball radius: {green_ball.radius}')
         print(f'Green ball angle: {green_ball.angle}\n')
 
     blue_ball = detected_balls.blue
-    if blue_ball.is_valid():
+    if blue_ball.found:
         print(f'Blue ball center: {blue_ball.center}')
         print(f'Blue ball radius: {blue_ball.radius}')
         print(f'Blue ball angle: {blue_ball.angle}\n')
--- a/examples/camera/camera_coloured_ball_detect.py
+++ b/examples/camera/camera_coloured_ball_detect.py
@@ -11,19 +11,22 @@ def process_frame(frame):
     if red_ball.found:
         print(f'Red ball center: {red_ball.center}')
         print(f'Red ball radius: {red_ball.radius}')
-        print(f'Red ball angle: {red_ball.angle}\n')
+        print(f'Red ball angle: {red_ball.angle}')
+        print()
 
     green_ball = detected_balls.green
     if green_ball.found:
         print(f'Green ball center: {green_ball.center}')
         print(f'Green ball radius: {green_ball.radius}')
-        print(f'Green ball angle: {green_ball.angle}\n')
+        print(f'Green ball angle: {green_ball.angle}')
+        print()
 
     blue_ball = detected_balls.blue
     if blue_ball.found:
         print(f'Blue ball center: {blue_ball.center}')
         print(f'Blue ball radius: {blue_ball.radius}')
-        print(f'Blue ball angle: {blue_ball.angle}\n')
+        print(f'Blue ball angle: {blue_ball.angle}')
+        print()
 
     cv2.imshow("Image", detected_balls.robot_view)
     cv2.waitKey(1)
--- a/examples/camera/camera_coloured_ball_detect.py
+++ b/examples/camera/camera_coloured_ball_detect.py
@@ -10,21 +10,27 @@ def process_frame(frame):
     red_ball = detected_balls.red
     if red_ball.is_valid():
         print(f'Red ball center: {red_ball.center}')
+        print(f'Red ball radius: {red_ball.radius}')
+        print(f'Red ball angle: {red_ball.angle}\n')
 
     green_ball = detected_balls.green
     if green_ball.is_valid():
         print(f'Green ball center: {green_ball.center}')
+        print(f'Green ball radius: {green_ball.radius}')
+        print(f'Green ball angle: {green_ball.angle}\n')
 
     blue_ball = detected_balls.blue
     if blue_ball.is_valid():
-        print(f'Blue ball center: {blue_ball.center}\n')
+        print(f'Blue ball center: {blue_ball.center}')
+        print(f'Blue ball radius: {blue_ball.radius}')
+        print(f'Blue ball angle: {blue_ball.angle}\n')
 
     cv2.imshow("Image", detected_balls.robot_view)
     cv2.waitKey(1)
 
 
 ball_detector = BallDetector()
-camera = Camera()
+camera = Camera(resolution=(640, 480))
 camera.on_frame = process_frame
 
 pause()
--- a/pitop/processing/algorithms/ball_detect.py
+++ b/pitop/processing/algorithms/ball_detect.py
@@ -255,7 +255,7 @@ class BallDetector:
 
         ball.center_points_cv.appendleft(ball_center_cv)
         ball.center = center_reposition(ball_center_cv, frame) if ball_center_cv is not None else None
-        ball.angle = get_object_target_lock_control_angle(ball.center, frame) if ball.is_valid is True else None
+        ball.angle = get_object_target_lock_control_angle(ball.center, frame) if ball_center_cv is not None else None
         ball.radius = ball_radius
 
         return ball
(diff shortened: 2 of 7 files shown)


examples/camera/camera_coloured_ball_detect.py
```diff
@@ -8,23 +8,29 @@ def process_frame(frame):
     detected_balls = ball_detector.detect(frame, color=["red", "green", "blue"])
 
     red_ball = detected_balls.red
-    if red_ball.is_valid():
+    if red_ball.found:
         print(f'Red ball center: {red_ball.center}')
+        print(f'Red ball radius: {red_ball.radius}')
+        print(f'Red ball angle: {red_ball.angle}\n')
```
m-roberts, line 14: Newline would be better as `print()` n next line.

pitop/processing/core/vision_functions.py
```diff
@@ -99,20 +99,21 @@ def get_object_target_lock_control_angle(center, frame):
...
+    from math import degrees
     if center is None:
         return None
     # physically, this represents an approximation between chassis rotation center and camera
     # the PID loop will deal with basically anything > 1 here, but Kp, Ki and Kd would need to change
     # with (0, 0) in the middle of the frame, it is currently set to be half the frame height below the frame
     chassis_center_y = -int(frame.shape[1])
 
-    # Anticlockwise is positive angle
+    # Clockwise is positive angle
```
m-roberts, line 113: Won't inverting the behaviour here affect the line follower?

duwudi: Just the comment was wrong - hence why comments are bad :D

examples/camera/camera_coloured_ball_detect.py
duwudi: you mean just add `print()` below?

tests/test_ball_detector.py
```diff
@@ -73,6 +79,12 @@ def test_detect_one_ball(self):
         # Check ball radii
         self.assertAlmostEqual(red_ball.radius, ball_radius, delta=self._MAX_DIMENSION_DIFFERENCE)
 
+        # Check angle
+        self.assertAlmostEqual(red_ball.angle, red_ball_angle, delta=self._MAX_DIMENSION_DIFFERENCE)
+
+        # Check center points deque has been appended
+        self.assertEqual(len(red_ball.center_points_cv), 1)
```
m-roberts, line 86: Why is it called `center_points_cv`? They are co-ordinates and have nothing to do with OpenCV?

pitop/processing/algorithms/ball_detect.py
```diff
@@ -36,40 +36,60 @@ def __init__(self, color):
         }
         self.match_limit = match_limits[color]
 
-        self.center_points = deque(maxlen=DETECTION_POINTS_BUFFER_LENGTH)
-        self.radius = 0
-        self.angle_from_center = None
+        self._center_points_cv = deque(maxlen=DETECTION_POINTS_BUFFER_LENGTH)
```
m-roberts, line 39: Why is it called `center_points_cv`? They are co-ordinates and have nothing to do with OpenCV?

pitop/processing/algorithms/ball_detect.py
```diff
@@ -36,40 +36,60 @@ def __init__(self, color):
...
+        self._center = None
+        self._radius = 0
+        self._angle = None
 
-    def clear(self):
-        self.center_points.clear()
+    def no_detection(self):
```
m-roberts, line 44: `ball.no_detection()` means nothing to me - it signals nothing of what it's going to do.

pitop/processing/algorithms/ball_detect.py
```diff
@@ -36,40 +36,60 @@ def __init__(self, color):
...
-        self.angle_from_center = None
+        self.angle = None
+
+    @property
+    def angle(self) -> float:
+        return self._angle
+
+    @angle.setter
+    def angle(self, value: float):
+        self._angle = value
```
m-roberts, line 56: What's the point?

pitop/processing/algorithms/ball_detect.py
```diff
@@ -36,40 +36,60 @@ def __init__(self, color):
...
+    @center.setter
+    def center(self, value: tuple):
+        self._center = value
+
+    @property
+    def radius(self) -> int:
+        return self._radius
+
+    @radius.setter
+    def radius(self, value: int):
```
m-roberts, line 71: Why use getters and setters like this if they behave the same way as a member variable?

pitop/processing/algorithms/ball_detect.py
duwudi: Our coordinate system for the user is referenced from the centre of the frame, so this is here to explicitly state these are OpenCV coordinates referenced from the top left of the frame

pitop/processing/algorithms/ball_detect.py
```diff
@@ -36,40 +36,60 @@ def __init__(self, color):
...
+    def found(self) -> bool:
         return self.center is not None
 
 
 class BallDetector:
     def __init__(self,
-                 process_image_width: int = 320,
-                 format: str = "OpenCV",
-                 print_fps: bool = False):
+                 image_processing_width: int = 320,
```
m-roberts, line 85: What if you want to specify its height? I think it would be better if you pass in a size tuple, with co-ordinate unset if you don't want to resize a particular axis. For example:
`image_processing_size=(320,)` would mean "resize to 320 width" as here, but is more flexible and more sanely named.

m-roberts: This will be a useful paradigm to get right

pitop/processing/algorithms/ball_detect.py
```diff
@@ -126,30 +148,28 @@ def __print_fps(self):
         print(f"[INFO] Approx. FPS: {self._fps.fps():.2f}")
 
     def __draw_ball_position(self, frame, ball):
-        self.cv2.circle(frame, ball.center, ball.radius, (0, 255, 255), 2)
-        self.cv2.circle(frame, ball.center, 5, tuple_for_color_by_name(ball.color, bgr=True), -1)
+        self.cv2.circle(frame, ball.center_points_cv[0], ball.radius, (0, 255, 255), 2)
```
m-roberts, line 151: Why did we drop `.center`?

pitop/processing/algorithms/ball_detect.py
duwudi: As I was interacting with it in the test file I thought it would be better to have a more defined interface, would also provide return value type checks for people interacting with the class in an IDE

pitop/processing/algorithms/ball_detect.py
duwudi: Well this would allow people to scale the width and height arbitrarily and we want it to be proportional. I see what you mean but I'd prefer to leave it as it is and just add a good description for the docs

examples/camera/camera_coloured_ball_detect.py
m-roberts:
```
print(f'Red ball angle: {red_ball.angle}')
print()
```

pitop/processing/algorithms/ball_detect.py
duwudi: Because it's in the wrong coordinates for OpenCV at this point in the code

pitop/processing/algorithms/ball_detect.py
duwudi: I'd be happier leaving this logic in the actual detect function tbh - I only changed it here because you added the clear() function. I had to change the clear function as it wasn't providing the intended functionality and hence I renamed it

pitop/processing/algorithms/ball_detect.py
m-roberts: Can you explain a little more about what you mean here? Perhaps a user code example?
My issue is more that it's not "opencv coordinates", top-left is standard for `(0,0)` in the SDK and elsewhere, so I would argue that this should just be `center_points` and wherever the user is getting 'centered/calibrated' co-ordinates should be made clear wherever they are used.

pitop/processing/algorithms/ball_detect.py
```diff
@@ -36,40 +36,60 @@ def __init__(self, color):
         }
         self.match_limit = match_limits[color]
 
-        self.center_points = deque(maxlen=DETECTION_POINTS_BUFFER_LENGTH)
-        self.radius = 0
-        self.angle_from_center = None
+        self._center_points_cv = deque(maxlen=DETECTION_POINTS_BUFFER_LENGTH)
+        self._center = None
```
m-roberts, line 40: What does this actually represent now? Before, this was "the current position of the center of the ball", which was a shorthand for "get latest value in deque". I'm not clear on what this is now and how it is used.

pitop/processing/algorithms/ball_detect.py
m-roberts: Okay - let's add docstrings then because - as I have highlighted in another comment - it's not entirely clear what they all represent.

pitop/processing/algorithms/ball_detect.py
m-roberts: It would, yes, but that would be on them if they did that. If that is not in our content, and the docs say "use width OR height to keep proportions the same, which you probably want" then it would be fine, and we can reuse this anywhere we need to scale an image for processing as it's generic.

pitop/processing/algorithms/ball_detect.py
duwudi: center is the user-facing center, where the center of the frame is (0, 0) - same as we use for the line detect stuff. It's way easier to write code for as a user if the middle of the frame is (0, 0) - it helps with face tracking, ball following, line following etc. The deque is a purely internal variable used for drawing the ball contrail

PR #334, opened 2021-04-22
Navigation controller

duwudi:
Support navigation through the `NavigationController` class. It inherits from `DriveController` so all of it's features are available through this new class.

The basic algorithm for navigation is, given an `(x, y)` position and an `angle`:
1. Turn robot to correct heading.
2. Drive to `(x, y)` position in a straight line.
3. Rotate to `angle`.

Also in this PR: 
- `DriveController` supports running it's motors synchronized. This feature is included in the latest expansion plate firmware, but it's backwards compatible with older versions. 
- Tuned wheel diameter and wheel separation in encoder motor and drive controller to be more accurate for dead reckoning position
- Changed the default parameter for `EncoderMotor` breaking type to be `BRAKE` rather than `COAST` - there is rarely any reason to be coasting for a robot if positional precision is desired

Currently, the `NavigatorController` only uses odometry data for pose estimation, but it has a Kalman filter implemented so that robot position tolerance can be given to user, e.g:
```
x = -0.49 m (+/- 0.0247)  # Best estimate for robot x is -0.49m but could be 0.0247m either side of that
y = 0.5 m (+/- 0.0247)
Angle = 1.79e+02° (+/- 5.92)
```

#### Example user code:

```
from pitop import DriveController, NavigationController
from pitop.pma import UltrasonicSensor

def do_somthing():
    # drop a ball off at location
    global goal_reached
    goal_reached = True

drive_controller = DriveController(left_motor_port="M3", right_motor_port="M0")
navigation_controller = NavigationController(drive_controller=drive_controller)
goal_reached = False

navigation_controller.go_to(position=(1, 0)).wait()  # program execution will block

navigation_controller.go_to(position=(1, 1), angle=60, on_finish=do_something)  # program will continue

while not goal_reached:
    if ultrasonic.distance < 0.1:
        navigation_controller.stop()
        print("You failed!")
        break
```
**_See example at examples/recipes/robot_navigate.py for more usage_**

#### Dependancies added

- `filterpy` relevant files were vendored.

#### Future Work

- Fuse IMU data into the Kalman filter to improve accuracy of dead reckoning
- Potentially: add ArUco markers for absolute pose positioning
- Add more advanced navigation methods that can follow a curved path (e.g. Pure pursuit path tracking)
- Add path planning


pitop/robotics/navigation/navigation_controller.py
```diff
@@ -0,0 +1,325 @@
...
+        self.angular_speed_factor = 0.0
+        self.max_v = 0.0
+        self.max_w = 0.0
+        self.deceleration_distance = 0.0
+        self.deceleration_angle = 0.0
+
+    def update_linear_speed(self, speed_factor):
+        self.linear_speed_factor = speed_factor
+        self.max_v = self.linear_speed_factor * self._max_motor_speed
+        self.deceleration_distance = self.linear_speed_factor * 0.4  # 0.4m at full speed
```
jcapona, line 67: I think we could set this `0.4` and `120` at the `update_angular_speed` method as class attributes

pitop/robotics/navigation/navigation_controller.py
```diff
@@ -0,0 +1,325 @@
...
+    def calc_distance(self, point_x, point_y):
+        dx = self.x - point_x
+        dy = self.y - point_y
+        return math.hypot(dx, dy)
+
+
+class GoalCriteria:
+    def __init__(self, linear_speed_factor, angular_speed_factor):
+        self._MAX_GOAL_REACHED_DISTANCE_ERROR = linear_speed_factor * 0.02  # 2 mm at full speed
+        self._MAX_GOAL_REACHED_ANGLE_ERROR = angular_speed_factor * math.radians(4.0)  # 4deg at full speed
```
jcapona, line 34: lets set the `0.02` and `4.0` used here as class attributes

jcapona (changes requested):
great stuff! suggested some small changes, mainly in variable names and setting constants you used directly as class attributes

examples/recipes/robot_navigate.py
```diff
@@ -0,0 +1,18 @@
+from pitop import DriveController, NavigationController
+
+drive_controller = DriveController(left_motor_port="M3", right_motor_port="M0")
+
+navigation_controller = NavigationController(drive_controller=drive_controller)
```
m-roberts, line 5: Why not have an internal drive controller? Can't see any use in declaring it outside...
```suggestion
from pitop import NavigationController

navigation_controller = NavigationController(left_motor_port="M3", right_motor_port="M0")
```

pitop/robotics/drive_controller.py
```diff
@@ -28,30 +28,33 @@ def __init__(self, left_motor_port="M3", right_motor_port="M0", name="drive"):
         self.right_motor_port = right_motor_port
         self.left_motor_port = left_motor_port
 
-        self._wheel_separation = 0.1675
-        self._wheel_diameter = 0.074
+        self._wheel_separation = 0.163
+        self._wheel_diameter = 0.0718
```
m-roberts, line 32: Can we avoid duplicating this? It's in the encoder motor as well

pitop/robotics/drive_controller.py
```diff
@@ -28,30 +28,33 @@ def __init__(self, left_motor_port="M3", right_motor_port="M0", name="drive"):
...
         self._wheel_circumference = self._wheel_diameter * pi
         self._linear_speed_x_hold = 0
 
         self._left_motor_port = left_motor_port
         self._right_motor_port = right_motor_port
 
         self.left_motor = EncoderMotor(port_name=left_motor_port,
-                                       forward_direction=ForwardDirection.CLOCKWISE)
+                                       forward_direction=ForwardDirection.CLOCKWISE,
+                                       wheel_diameter=self._wheel_diameter)
```
m-roberts, line 41: Not needed if a constant value is always passed in that matches the parameter's default value

pitop/robotics/navigation/navigation_controller.py
```diff
@@ -0,0 +1,323 @@
...
+from dataclasses import dataclass
+import numpy as np
+import math
+from typing import Union
+from simple_pid import PID
+import sched
+
+
+@dataclass
+class RobotState:
```
m-roberts, line 12: A bit vaguely named

pitop/robotics/navigation/navigation_controller.py
```diff
@@ -0,0 +1,323 @@
...
+    x: float = 0.0
+    y: float = 0.0
+    theta: float = 0.0
+    v: float = 0.0
+    w: float = 0.0
+
+    @property
+    def pose(self):
+        return self.x, self.y, self.theta
```
m-roberts, line 21: Not sure about returning multiple values like this. What is a 'pose', anyway?

pitop/robotics/navigation/navigation_controller.py
```diff
@@ -276,7 +323,7 @@ def __sub_goal_reached(self):
 
     def __get_new_pose_update(self):
         self._position_update_event.wait()
-        return self.robot_state.pose
+        return self.robot_state.x, self.robot_state.y, self.robot_state.theta
```
m-roberts, line 324: Do we need to return 3 parameters? Or can we group them somehow?

pitop/robotics/navigation/navigation_controller.py
```diff
@@ -0,0 +1,368 @@
...
+        return math.hypot(dx, dy)
+
+
+class GoalCriteria:
+    def __init__(self, linear_speed_factor, angular_speed_factor):
+        self._MAX_GOAL_REACHED_DISTANCE_ERROR = linear_speed_factor * 0.02  # 2 mm at full speed
+        self._MAX_GOAL_REACHED_ANGLE_ERROR = angular_speed_factor * math.radians(4.0)  # 4deg at full speed
+
+    def angle(self, angle_error):
+        return abs(angle_error) < self._MAX_GOAL_REACHED_ANGLE_ERROR:
```
m-roberts, line 83:
```suggestion
        return abs(angle_error) < self._MAX_GOAL_REACHED_ANGLE_ERROR
```

m-roberts: How did this get past your syntax checking?! 😝

pitop/robotics/navigation/navigation_controller.py
```diff
@@ -0,0 +1,325 @@
...
+
+    def distance(self, distance_error, heading_error):
+        if abs(distance_error) < self._MAX_GOAL_REACHED_DISTANCE_ERROR:
+            return True
+        if abs(heading_error) > math.pi / 2:
+            # Overshot goal, error will be small so consider goal to be reached.
+            # Otherwise robot would have to reverse a tiny amount which would be poor UX.
+            # Robot will correct itself on subsequent navigation calls anyway.
+            return True
+        return False
```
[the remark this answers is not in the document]
m-roberts: I think I prefer the readability of how it is now.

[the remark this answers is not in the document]
duwudi: Yeah I left it in two separate if statements to make sure we could comment on that condition - it's a bit of an odd one to understand

examples/recipes/robot_navigate.py
duwudi: we're using the drive controller inside the navigation controller - I'd rather not duplicate that code. 

Also, there are many situations where a user would want to add both a drive controller and a navigation controller to their robot. E.g. use `robot.drive.target_lock_drive_angle()` to go and get a coloured ball, then they would switch to navigation controller to go and drop it off at a certain location. In this case we want to pass in DriveController directly so that they don't have to duplicate port assignment

pitop/robotics/navigation/navigation_controller.py
duwudi: I'm not sure how to make this more clear, it's the full state of the robot - this includes all degrees of freedom for this robot and their derivates. Any suggestions?

pitop/robotics/navigation/navigation_controller.py
duwudi: pose is every variable required to estimate a robot's position and orientation in space. In our case, we're on a 2D plane so just x, y and theta is required. I've deleted this anyway as I was getting in a mess with angles (giving user degrees and having radians internally) - I've just changed to `position` that gives x and y as this is consistent with the `go_to` arguments of `position` and `angle`

pitop/robotics/navigation/navigation_controller.py
duwudi: `pose` grouping is what makes the most sense, but I've since changed this because of the other comment above

pitop/robotics/navigation/navigation_controller.py
duwudi: haha I think it was Jorge's suggested commit that I confirmed inside the GitHub web page so didn't catch it

pitop/robotics/navigation/navigation_controller.py
```diff
@@ -0,0 +1,407 @@
...
+
+    def update_linear_speed(self, speed_factor):
+        self._max_distance_error = speed_factor * self._full_speed_distance_error
+
+    def update_angular_speed(self, speed_factor):
+        self._max_angle_error = speed_factor * self._full_speed_angle_error
+
+
+class RobotDrivingManager:
```
duwudi, line 106: Name suggestions welcome, currently a bit ambiguous as to `drive_controller` vs `drive_manager`

m-roberts:
CI is failing:
```
W: debian-changelog-line-too-long line 35
W: debian-changelog-line-too-long line 47
W: debian-changelog-line-too-long line 80
W: debian-changelog-line-too-long line 86
W: spelling-error-in-changelog dependancy dependency
W: spelling-error-in-changelog refector refactor
```
These need to be filtered out for the build to pass

PR #335, opened 2021-04-23
Configurable web interface API - refactor proposal

m-roberts:


olivierwilkinson:
I think we need a meeting to discuss most of this, looking through the changes here I feel there is a misunderstanding of quite of few decisions made in the previous PR.

PR #342, opened 2021-05-07
Drive Controller to use speeds instead of RPMs

duwudi:
Use speed values inside drive controller instead of using RPM values (small thing but it's been bothering me for a while)

- Makes calculations easier 
- Doesn't duplicate code already inside encoder motor for doing conversions
- Keeps RPMs away from higher-level classes completely, and leaves that solely to encoder motor class


m-roberts:
Seems like there is an issue with the tests. It seems to be related to mocks

duwudi:
> Seems like there is an issue with the tests. It seems to be related to mocks

Not sure why this was passing locally for me on Friday (definitely shouldn't have been), but there was an issue with the max speed calculation

jcapona (approved):

PR #344, opened 2021-05-10
Remove double Pitop __init__ in AlexRobot

duwudi:
```
from pitop import AlexRobot
alex = AlexRobot()
```
This hangs presumably because the Miniscreen is already taken from the `Pitop.from_config` command already run


m-roberts:
I don't understand where the "double" is, nor what you mean by
> Miniscreen is already taken from the Pitop.from_config command already run

I am also not a fan of "I'm not sure if this is the correct way to fix this, but it works" - root cause needs more investigation before we commit to a fix that we don't understand.

PR #352, opened 2021-05-25
Add optional properties for autocomplete functionality when doing robotics

duwudi:
Putting this here for discussion purposes mainly, we got [this forum](https://forum.pi-top.com/t/nothing-happens-when-calibrating-motors-on-alex/844/6?u=duwudi) question recently and it reminded me of the discussion we had before - this PR implements @angusjfw's suggestion from that.

These are the limitations with the approach I can see (there are probably more):

1. There are a lot more components that can be added to a Pitop object dynamically and this will grow with time (e.g. PMA, NavigationController etc) - this currently only covers the common properties for robotics. Could we add these Optional properties based on a master list we keep somewhere that defines all `Recreatable` components?
2. People might incorrectly assume that these properties are available before they've used the `add_component` or `from_config` methods - resulting in errors and confusion. I'm not sure if we can catch this at runtime and provide an explanation of what to do but that would certainly help. My hope is that they will follow the examples properly and in the cases where they don't, will read the docs and see how to do it properly.

### DIFF
--- a/pitop/system/pitop.py
+++ b/pitop/system/pitop.py
@@ -4,6 +4,8 @@ from pitop.core.mixins import (
     SupportsBattery,
     SupportsMiniscreen,
 )
+from typing import Optional
+from pitop import DriveController, Camera, PanTiltController
 
 
 class Pitop(SupportsMiniscreen, SupportsBattery, Componentable, metaclass=Singleton):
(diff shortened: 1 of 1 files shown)


duwudi:
Don't think there is a very clean way of achieving this with the current architecture, closing for now

Issue #367, opened 2021-06-09
Camera() instantiation breaks stable usage of UltrasonicSensor()


duwudi:
```
from pitop import (
    Camera,
    UltrasonicSensor,
)
from time import sleep

camera = Camera(resolution=(640, 480))

ultrasonic = UltrasonicSensor("D3")

while True:
    print(f"Distance: {ultrasonic.distance}")
    sleep(0.1)
```

If the camera instantiation is commented out, the program runs normally and distance readings make sense.

With the camera, the following problems occur:
- The first call to `ultrasonic.distance` hangs for 8-10 seconds. This hanging also happens sporadically throughout subsequent calls too.
- The readings returned are unstable and do not represent the actual distance at all. For example, the distance returned with the hand at around 0.1m away from the sensor is 0.6m - the returned readings seem to span 0.6-1.0m regardless of the true distance.

htop screenshot whilst program running:

<img width="521" alt="2021-06-09 14_15_51-192 168 1 153 (pi-top) - VNC Viewer" src="https://user-images.githubusercontent.com/58126123/121303109-413a8e80-c92d-11eb-8b39-d3d2922e37c7.png">

duwudi:
Note: workaround for this is https://github.com/pi-top/pi-top-Python-SDK/pull/368 - we should encourage using A1 and A3 for the ultrasonic sensor as it isn't susceptible to timing issues that are prevalent on the Raspberry Pi

PR #368, opened 2021-06-09
Add MCU-based Ultrasonic Sensor readings on ports A1 and A3

duwudi:
- Converted UltrasonicSensor class to be able to choose from different UltrasonicSensor implementations (`RPI` or `MCU`) based on the port given by the user. 
- Note: factory method would have been preferred but this option left generated documentation the cleanest.
- Added an ultrasonic sensor base class to avoid duplication of identical methods across both implementations

## **To do**

1. ~~Implement `when_activated`, `when_deactivated`, `wait_for_active` and `wait_for_inactive` for MCU mode. Will need to add threading and decide how often to read the ultrasonic sensor, not sure how GPIOZero implements this~~
2. ~~The MCU currently retains the last value that was read from the ultrasonic sensor even if the sensor has been unplugged. Ideally, we want it to revert back to default state~~ - **can't do without firmware change, skip for now**
3. ~~Behaviour of RPI ultrasonic readings is to return the max distance if the reading fails (e.g. if the ultrasonic is not plugged into the correct port), the MCU returns 0.0 in this case so we should probably catch that and return max_distance instead.~~
4. ~~Add expansion plate firmware version check on initialization~~

### **Bizarre problem discovered _(for reference of code, now fixed)_** 
Whilst testing ultrasonic with FaceDetector class, I found they were incompatible if the FaceDetector was instantiated after the UltrasonicSensor on an analog port. Tracked it down to the loading of the dlib shape predictor causing the issue - see comments in code below:
```
from pitop.pma.ultrasonic_sensor import UltrasonicSensor
from time import sleep
import dlib


ultrasonic = UltrasonicSensor("A1")

# This takes a long time to load and blocks the heartbeat signal from sending to EP
# Causing the EP's ultrasonic sensor configuration to reset (along with all other configs) 
# distance readings become 0, and any other commands expecting a specific configuration will also not behave as expected
predictor = dlib.shape_predictor("shape_predictor_68_face_landmarks.dat")

while True:
    print(f"Distance: {ultrasonic.distance}")  # This would now be 0.0 since A1 configuration is reset 
    sleep(0.1)
```
To fix, heartbeat signal shutdown time can be increased as per d803aacad10bd8611b1cbde8f90ccd39a8912e48

Other options:
- Look into using a scheduler for the heartbeat signal instead of while True with sleep, since `sched` has a priority level it may still work even whilst loading the dlib shape predictor.
- Load shape predictor in a separate thread - this may not actually work since the heartbeat and shape predictor loading is still happening in two distinct threads anyway. Also a thread that loads the predictor may cause timing issues with the rest of the program.


pitop/pma/ultrasonic_sensor_base.py
```diff
@@ -0,0 +1,347 @@
...
+    def threshold_distance(self, value):
+        self.threshold = value / self.max_distance
+
+    @property
+    def distance(self):
+        return self.value * self._max_distance
+
+    @abstractmethod
+    def in_range(self):
+        pass
```
m-roberts, line 54:
```suggestion
        raise NotImplementedError
```

m-roberts: If it's being inherited and is required, it should raise an exception when it hasn't been explicitly overridden by the subclass and it is called.

m-roberts: (same for all `pass`es)

m-roberts:
Increasing timeout to "solve" a hanging thread issue seems like it's ignoring the fundamental issue. However, if it's strictly internal, then I suppose we can release that and create a new Issue for addressing the underlying problem.

Finding the diff very hard to read - can you add example code that shows how the user actually handles this, and how the user code is different when using a factory method? Would also like to know what the "unclean" factory method docs look like so we can make an informed decision.

Is this a breaking change for the SDK?

pitop/pma/ultrasonic_sensor.py
```diff
@@ -1,61 +1,47 @@
...
+        if port_name in valid_analog_ports:
+            self.__ultrasonic_device = UltrasonicSensorMCU(port_name=port_name,
+                                                           queue_len=queue_len,
+                                                           max_distance=max_distance,
+                                                           threshold_distance=threshold_distance,
+                                                           partial=partial,
+                                                           name=name
+                                                           )
+        else:
+            self.__ultrasonic_device = UltrasonicSensorRPI(port_name=port_name,
```
angusjfw, line 35: As an alternative to this or a factory method, I think it would be possible to dynamically inherit from these classes with the use of a [metaclass ](https://stackoverflow.com/a/100146). On the other hand this is the pattern used to manage the multiple implementations of the Camera component also and is less confusing.

pitop/pma/ultrasonic_sensor.py
```diff
@@ -1,61 +1,47 @@
...
-            self._echo_rise = None
-            self._echo_fall = None
-            self.pin.edges = 'both'
-            self.pin.bounce = None
-            self.pin.when_changed = self._echo_changed
-            self._queue.start()
-        except Exception:
-            self.close()
-            raise
+        if port_name in valid_analog_ports:
```
angusjfw, line 26: I think there should be a more explicit api for selecting these implementations than passing a particular port number, along the lines of CameraType. Also this blocks using ports A1 or A3 for UltrasonicSensorRPI which I don't think is necessary?

pitop/pma/ultrasonic_sensor.py
```diff
@@ -65,56 +51,18 @@ def __init__(
                                     "max_distance": lambda: self.max_distance,
                                     "threshold_distance": lambda: self.threshold_distance})
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.close()
```
angusjfw, line 58: we should add an example of using this to switch the use of a port

duwudi:
> Increasing timeout to "solve" a hanging thread issue seems like it's ignoring the fundamental issue. However, if it's strictly internal, then I suppose we can release that and create a new Issue for addressing the underlying problem.

Not sure how to solve the underlying issue, but yeah we can create a ticket for it.

> Finding the diff very hard to read - can you add example code that shows how the user actually handles this, and how the user code is different when using a factory method? Would also like to know what the "unclean" factory method docs look like so we can make an informed decision.

User doesn't need to do anything differently, it's exactly the same from their perspective. For the factory method, they would be returned the correct class based on their input so again the user code would be no different. I don't know how to deal with documentation for that, I just followed the same way we did this for the camera which seemed like the least friction approach.

> Is this a breaking change for the SDK?
Shouldn't be no

pitop/pma/ultrasonic_sensor.py
duwudi: UltrasonicSensorRPi = controlled via RPi digital pins
UltrasonicSensorMCU = controlled via expansion plate MCU pins and accessed over I2C by Raspberry Pi

UltrasonicSensorRPi can't use A1 and A3 so I don't think there is a need for a user to choose an implementation through an API as it depends solely on which port they have it connected to. It would be nice if the pros and cons of each were described, but in general, I think for all robotics Further content we should be using UltrasonicSensorMCU since it's more robust (which means telling user to plug into A1 or A3)

pitop/pma/ultrasonic_sensor_base.py
duwudi: yep!

angusjfw:
> To fix, heartbeat signal shutdown time can be increased as per [d803aac](https://github.com/pi-top/pi-top-Python-SDK/commit/d803aacad10bd8611b1cbde8f90ccd39a8912e48)
> 
> Other options:
> 
> * Look into using a scheduler for the heartbeat signal instead of while True with sleep, since `sched` has a priority level it may still work even whilst loading the dlib shape predictor.
> * Load shape predictor in a separate thread - this may not actually work since the heartbeat and shape predictor loading is still happening in two distinct threads anyway. Also a thread that loads the predictor may cause timing issues with the rest of the program.

It's likely this problem cannot be fixed with threading in python due to the GIL, if file is being read for several seconds no other thread can execute during this time. Asyncio loop may be able to continue while this is happening but would be a lot of work to add alongside threading in our sdk. 

I think a simple, viable solution might be to run the heartbeat loop as a subprocess, which can run concurrently and only has the downside of being awkward to extract output which shouldn't matter in that use case.

m-roberts:
Let's split this out. The timing issue is totally separate from the intention of this PR and it's muddying the conversation.

pitop/pma/ultrasonic_sensor.py
duwudi: Possibly yeah, I read [this](https://realpython.com/factory-method-python/) when researching how to do this - weird they don't mention using a metaclass as it does seem ideal. I also can't find many examples of using a metaclass for this purpose but I'll definitely try to learn more about it. The dynamic inheritance does bother me in terms of documentation and general readability for users though, but if this was being done from scratch (as opposed to a bolt-on fix) I'd probably explore these more exotic options

pitop/pma/ultrasonic_sensor.py
duwudi: Yeah I think gpiozero have some explanations on that too, only really useful if you're using python interactively though

duwudi:
> I think a simple, viable solution might be to run the heartbeat loop as a subprocess, which can run concurrently and only has the downside of being awkward to extract output which shouldn't matter in that use case.

That's a pretty good idea! We'll need this heartbeat time extension for now anyway, but can revert it back once we have a better solution

pitop/pma/ultrasonic_sensor.py
olivierwilkinson: I agree a metaclass would be a good way to do this, it would also remove a lot of the code related to accessing properties on the underlying Ultrasonic device and enforce that the base classes need the same API. @PiTopRyan there is a really good overview on stackoverflow [here](https://stackoverflow.com/a/6581949)... someone reallllllllly wanted those upvotes 😆.

pitop/pma/ultrasonic_sensor.py
angusjfw: Right yeah, confused myself. In that case this does seem like a good way of selecting the implementation, and we'll have to explain the differences and constraints in the docs or elsewhere.

pitop/pma/ultrasonic_sensor.py
duwudi: > Metaclasses are deeper magic that 99% of users should never worry about it. If you wonder whether you need them, you don't (the people who actually need them to know with certainty that they need them and don't need an explanation about why).

Classic 😄 Great explanation on that post though!

angusjfw (commented):
I've tested the new MCU implementation alongside the existing RPI one and can confirm it fixes #367. It also does appear to produce slightly more accurate readings, particularly given that the rpi implementation only returns multiples of 0.3, and it responds faster to changes.

angusjfw:
I think we need to resolve what happens when you try to instantiate ultrasonics on both ports A1 and A3 at the same time.
As I expected, this generally didn't seem to work and I would get errors related to the device id or firmware version, usually this:
```
  File "ultra.py", line 11, in <module>
    ultrasonic1 = UltrasonicSensor("A3")
  File "/home/pi/pi-top-Python-SDK/pitop/pma/ultrasonic_sensor.py", line 32, in __init__
    name=name
  File "/home/pi/pi-top-Python-SDK/pitop/pma/ultrasonic_sensor_base.py", line 104, in __init__
    self.__compatibility_check()
  File "/home/pi/pi-top-Python-SDK/pitop/pma/ultrasonic_sensor_base.py", line 233, in __compatibility_check
    raise RuntimeError("You are trying to use analog ports for the Ultrasonic Sensor, this is not compatible "
RuntimeError: You are trying to use analog ports for the Ultrasonic Sensor, this is not compatible with a Foundation Plate. Please use an Expansion Plate instead.
```

And occasionally this:
```
pi@pi-top:~/pi-top-Python-SDK $ python3 ultra2.py
  File "ultra2.py", line 10, in <module>
    ultrasonic = UltrasonicSensor("A3")
  File "/home/pi/pi-top-Python-SDK/pitop/pma/ultrasonic_sensor.py", line 32, in __init__
    name=name
  File "/home/pi/pi-top-Python-SDK/pitop/pma/ultrasonic_sensor_base.py", line 104, in __init__
    self.__compatibility_check()
  File "/home/pi/pi-top-Python-SDK/pitop/pma/ultrasonic_sensor_base.py", line 236, in __compatibility_check
    firmware_device = FirmwareDevice(plate_id)
  File "/usr/lib/python3/dist-packages/pitopcommon/firmware_device.py", line 80, in __init__
    "Part name provided does not match. {} != {}".format(hex(self.part_name), hex(self.get_part_name())))
pitopcommon.firmware_device.PTInvalidFirmwareDeviceException: Part name provided does not match. 0x2222 != 0x2222
```

However, I did actually find that I was sometimes able to get correct readings from both ports without error, when instantiating the sensors in two separate python scripts run separately at the command line. So perhaps the hardware/firmware does allow for this? I only have one sensor at the moment so I was swapping it between the ports, and found the reading would stay unchanged for the device that was not plugged in until I replaced it, then it would work again.

When I was getting these errors I run `pi-top devices` to confirm I was not having issues communicating with the hub/expansion plate and it all seemed ok.

I think we need to either make this work reliably if possible, or improve the error messages, validation and docs to explain that this is not possible.

duwudi:
> I would get errors related to the device id or firmware version

I've tracked this down to an incompatibility between the two different i2c methods we have in the SDK. The legacy I2CDevice class (which is non-standard I2C that somehow made into the SDK a long time ago) is used for any registers related to firmware updating (version, part number etc) and hub communication yet we use normal SMBus for interacting with plates. 

Since the first instantiated ultrasonic sensor is already reading data at 10 Hz it appears to corrupt the second ultrasonic compatibility check that uses the `I2CDevice` class (inside `FirmwareDevice()`). I confirmed this by increasing the time between consecutive reads in the ultrasonic read loop, when I set it to 1 second as opposed to 0.1 seconds it seems to work fine - i.e. the part names and version numbers are not corrupted.

Interestingly, I did a test by running `pi-top devices` whilst I was running a single ultrasonic sensor on loop in another terminal - this seems to work fine - I expected this to fail in the same way but it doesn't seem to be impacted. So perhaps running in the same process is the problem, but in separate processes it's fine - this is consistent with what you found too.

duwudi:
> So perhaps running in the same process is the problem

Just put the compatibility check into its own process and it works fine now, I'd prefer to try and get closer to the root cause before implementing this though

angusjfw:
> @angusjfw try out [908dae9](https://github.com/pi-top/pi-top-Python-SDK/commit/908dae9d3a1c115f1f7aa977e55732741522f0e6) and let me know what you think

Yeah, this seems to work reliably now when creating two sensors in one script. When running two scripts it's often fine but can still fail, I assume with the same issue that the data reads corrupt the compatibility check.

This seems reasonable to me for now, we can try to fix the legacy I2CDevice issue later.

duwudi:
> When running two scripts it's often fine but can still fail, I assume with the same issue that the data reads corrupt the compatibility check.

That's annoying, I thought it worked fine in separate processes. I just tried the same thing and yeah it's giving me the "no expansion plate attached" error 4/5 times. The weird thing is that running `pi-top devices` in the terminal never fails regardless of what's running, it always says that I have an Expansion Plate attached - need to dig into why that isn't failing yet the second ultrasonic sensor is

> This seems reasonable to me for now, we can try to fix the legacy I2CDevice issue later.

I wish we could just change all the firmware to use SMBus but we're kind of stuck with it now.

duwudi:
> The weird thing is that running `pi-top devices` in the terminal never fails regardless of what's running, it always says that I have an Expansion Plate attached - need to dig into why that isn't failing yet the second ultrasonic sensor is

ok it does fail sometimes, perhaps we just need to implement a global lock on any i2c transactions regardless of whether it's the `I2CDevice` or `SMBusDevice`. Kind of a big change though and I'm not sure of the overhead of doing that for every single transaction. To be honest I thought this stuff would be taken care of on the kernel level, clearly not

angusjfw (dismissed):

pitop/pma/ultrasonic_sensor.py
olivierwilkinson: Just to clarify, I think a metaclass would be appropriate here but it wouldn't make a big enough difference to prevent merging 😄 Maybe a comment in the ultrasonic_sensor_base file saying the two underlying classes should be kept as similar as possible to this class would be good, but even then only if you can be bothered.

m-roberts:
just fixed references to `pitopcommon` that were causing this PR to fail. is there any reason not to merge this in?

duwudi:
> just fixed references to `pitopcommon` that were causing this PR to fail. is there any reason not to merge this in?

Thanks! Angus and Liv are happy with it, so if you could merge that would be great 👍

PR #369, opened 2021-06-09
Use multiprocessing.Process on Camera class

### LINKED ISSUE #367: Camera() instantiation breaks stable usage of UltrasonicSensor()

jcapona:
Closes #367 

Use `multiprocessing.Process` instead of `threading.Thread` in `Camera` class.

### DIFF
--- a/pitop/camera/camera.py
+++ b/pitop/camera/camera.py
@@ -14,8 +14,7 @@ from pitop.pma.common import type_check
 
 from enum import Enum
 from inspect import signature
-from multiprocessing import Process
-from threading import Event
+from multiprocessing import Event, Process
 
 
 class Camera(Stateful, Recreatable):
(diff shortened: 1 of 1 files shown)


m-roberts (approved):

m-roberts:
@jcapona has reported that this is causing issues with running Python camera examples from scripts.

angusjfw (changes requested):
This doesn't work because the frames are now being processed by a separate instance of Camera in the forked process. There are a several methods of sending or sharing objects between processes provided by [multiprocessing](https://docs.python.org/3/library/multiprocessing.html#exchanging-objects-between-processes).

PR #371, opened 2021-06-10
Camera with Ultrasonic tests

### LINKED ISSUE #367: Camera() instantiation breaks stable usage of UltrasonicSensor()

duwudi:
Adding sleep to threading loop fixes #367

Also added a demo example of using multiprocessing to get frames from the camera instead of a thread, also fixes #367

**For discussion only, haven't tested the implications of doing this across other use-cases**

### DIFF
--- a/pitop/camera/camera.py
+++ b/pitop/camera/camera.py
@@ -16,6 +16,8 @@ from enum import Enum
 from inspect import signature
 from threading import Thread, Event
 
+from time import sleep
+
 
 class Camera(Stateful, Recreatable):
     """Provides a variety of high-level functionality for using the PMA USB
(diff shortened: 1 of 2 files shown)


pitop/camera/camera.py
```diff
@@ -280,6 +282,7 @@ def __process_camera_output(self):
                 self.__frame_handler.process()
             except Exception as e:
                 print(f"Error in camera frame handler: {e}")
+            sleep(0.015)
```
markpitop, line 285: I'm really not sure this is a good idea, but regardless, you definitely want to add a detailed comment about what it's doing and alternative approaches considered.

duwudi: yeah it wasn't meant to be a good idea, just for discussion to help diagnosis

duwudi:
Closing since the ultrasonic problem is avoided by https://github.com/pi-top/pi-top-Python-SDK/pull/368

PR #394, opened 2021-07-22
Prax head shake and nod

duwudi:
Required for Robotics 102 courses

Best that this is implemented in the SDK as opposed to the user coding it as it would distract too much, also having a non-blocking version requires threads which we definitely don't want users doing

**One of the examples requires https://github.com/pi-top/pi-top-Python-SDK/pull/368 to be merged as it uses A1 for Ultrasonic**


pitop/robotics/tilt_roll_head_controller.py
```diff
@@ -25,32 +27,89 @@ def __init__(self, servo_roll_port="S0", servo_tilt_port="S3", name="head"):
...
+        """
+        self.__start_servo_oscillation(servo=self._tilt_servo, times=times, angle=angle, speed=speed, block=block)
+
+    def __start_servo_oscillation(self, servo: ServoMotor, times: int, angle: int, speed: int, block: bool) -> None:
+        if block:
+            self.__oscillate_servo(servo=servo, times=times, angle=angle, speed=speed)
+        else:
+            Thread(target=self.__oscillate_servo,
+                   kwargs={"servo": servo, "times": times, "angle": angle, "speed": speed},
+                   daemon=True).start()
```
jcapona, line 77: what should be the behavior when 2 non-blocking movements are requested sequentially? my guess would be that the last movement is the one that gets executed...
anyway, I think we should limit the amount of threads the class can create and keep track of it, so you can "kill it" if this scenario comes up!

jcapona (changes requested):

pitop/robotics/tilt_roll_head_controller.py
duwudi: @jcapona good catch - I've implemented this now let me know what you think

Issue #376, opened 2021-06-16
Add text-to-speech functionality


duwudi:
Use pi-top's onboard speaker to interact with the user via speech - particularly useful/interesting for robotics applications but can also have a lot of value for simpler Foundation Kit projects.

See old library created [here ](https://github.com/pi-top/python3-pt-speech-deb)that uses `espeak`. 

- Possibly make this part of the `Pitop` class so it can be used by default without the user specifically adding it manually
- Would be nice to have multiple implementations available to use, i.e. user can choose between a high-quality cloud-based version (Google/Amazon/Microsoft) or a lower quality local version (espeak etc)

angusjfw:
I got pretty good results using festival with a plug-in voice as described [here](https://askubuntu.com/a/908889).

```
sudo apt-get install -y festival festvox-us-slt-hts
festival -b '(voice_cmu_us_slt_arctic_hts)' \
    '(SayText "The temperature is 22 degrees centigrade and there is a slight breeze from the west.")'
```

That python lib actually binds to the c++ lib, subprocesses are only used for converting to mp3 I think. It gives me an ImportError when I try to use it though.

angusjfw:
Similar results with flite:
```
sudo apt install flite
flite -voice slt -t "The Raspberry Pi is a great Maker platform!"
```

duwudi:
> I got pretty good results using festival with a plug-in voice as described [here](https://askubuntu.com/a/908889).

This one sounds pretty good actually, let's just use that!

duwudi:
> The [python lib](https://github.com/techiaith/pyfestival) actually binds to the c++ lib, subprocesses are only used for converting to mp3 I think. It gives me an ImportError when I try to use it though.

Can we still use the plugin if using the python library?

duwudi:
> It gives me an ImportError when I try to use it though.

Yeah I've got the same thing, the pypy package is not actually up to date with the repo though - importing from the pip-installed version gives this error:

```
Traceback (most recent call last):
  File "/home/pi/.local/lib/python3.7/site-packages/festival.py", line 3, in <module>
    from . import _festival
ImportError: attempted relative import with no known parent package
```

Installing from the repo directly gives this error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/pi/.local/lib/python3.7/site-packages/festival.py", line 6, in <module>
    import _festival
[...]

PR #405, opened 2021-07-28
Text to speech mixin

duwudi:
No implementation yet, just a wireframe and example usage

When finished, will close https://github.com/pi-top/pi-top-Python-SDK/issues/376

### DIFF
--- a/pitop/processing/speech/text_to_speech.py
+++ b/pitop/processing/speech/text_to_speech.py
@@ -1,0 +1,22 @@
+class TTS:
+    __VOICES = ()
+
+    def __init__(self):
+        self._voice = None
+
+    def __call__(self, text: str):
+        print(text)
+
+    @property
+    def voice(self):
+        return self._voice
+
+    @voice.setter
+    def voice(self, voice: str):
+        if voice not in self.__VOICES:
+            raise ValueError(f"Invalid voice choice. "
+                             f"Choose from {', '.join(self.__VOICES[:-1])} or {self.__VOICES[-1]}")
+        self._voice = voice
+
+    def list_voices(self):
+        print(f"Available voices: f{self.__VOICES}")
--- a/examples/system/pitop_speech.py
+++ b/examples/system/pitop_speech.py
@@ -1,0 +1,7 @@
+from pitop import Pitop
+
+pitop = Pitop()
+
+while True:
+    text = input("Enter text: ")
+    pitop.speak(text)
--- a/pitop/core/mixins/__init__.py
+++ b/pitop/core/mixins/__init__.py
@@ -3,3 +3,4 @@ from .recreatable import Recreatable
 from .stateful import Stateful
 from .supports_battery import SupportsBattery
 from .supports_miniscreen import SupportsMiniscreen
+from .supports_speech import SupportsSpeech
--- a/pitop/core/mixins/supports_speech.py
+++ b/pitop/core/mixins/supports_speech.py
@@ -1,19 +1,8 @@
-from pitop.core.exceptions import UnavailableComponent
-from pitop.system import device_type
-
-from pitopcommon.common_names import DeviceName
-
-
 class SupportsSpeech:
     def __init__(self):
         from pitop.processing.speech import TTS
-        self._tts = None
-        if device_type() == DeviceName.pi_top_4.value:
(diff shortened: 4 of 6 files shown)


pitop/processing/speech/text_to_speech.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __init__(self):
+        self._speed = 1.0
+        self.__voices = self.get_voices()
+        self._language = None
+        self.language = "us"
+        self._voice = None
+        self.voice = self.__voices.get(self.language)[0]
+
+    def __call__(self, text: str):
```
angusjfw, line 16: I think we should make this available as a `say` method also, particularly for use of the TTS class as a standalone without creating a Pitop object.

debian/py3dist-overrides
```diff
@@ -21,3 +21,4 @@ spidev python3-spidev; PEP386
 systemd_python python3-systemd; PEP386
 wget python3-wget; PEP386
 pyzmq python3-zmq; PEP386
+pyfestival python3-pt-pyfestival
```
m-roberts, line 24:
```suggestion
pyfestival python3-pyfestival; PEP386
```

debian/control
```diff
@@ -82,6 +82,9 @@ Replaces:
  pt-device-manager (<< 4.0.0),
 # pt-oled
  python3-pt-oled (<< 3.0.0),
+# Festival speech engine
```
m-roberts, line 85: This is currently under `Replaces:` field, which is [definitely not what you want to be doing](https://www.debian.org/doc/debian-policy/ch-relationships.html#overwriting-files-and-replacing-packages-replaces).

Are you trying to make this a core dependency, an optional dependency or a dependency of the full SDK?

debian/py3dist-overrides
m-roberts: `PEP386`  flag will make the dependency versioned, which we want in most cases.

setup.py
```diff
@@ -137,6 +137,11 @@
     # Download Model Files #
     ########################
     "wget>=3.2,<4.0",
+
+    ##################
+    # Text to Speech #
+    ##################
+    "pt-pyfestival",
```
m-roberts, line 144:
```suggestion
    "pyfestival",
```

m-roberts: This will not install correctly for people installing via Python on RPi for now, but we want tackle that later on.

m-roberts:
I am not convinced with the design pattern. Why should the user need to know anything about festival? Why does the user need to handle the importing and management of TTS directly?

```
from pitop import Pitop

pitop = Pitop()
pitop.speak("Hello World")
# Another option might be a 'speaker' property of the pi-top:
# pitop.speaker.speak("Hello World")
```
seems like a much more sensible way to approach this.

m-roberts:
This is currently failing to build the package as there is a spelling mistake in [this commit](https://github.com/pi-top/pi-top-Python-SDK/pull/405/commits/30bc74961d6d993e1c5d92cf430aa27f9d64927d)'s message that is propagating into the snapshot version changelog.

[This draft PR](https://github.com/pi-top/debian-package-build-action/pull/19) provides an easy way for all package builds to ignore changelog spelling mistakes for snapshot builds with a simple boolean input. This hasn't been tested, and this PR is a good test of this functionality.

We should update this PR to use this experimental branch until it passes, then merge in and build against this new master commit. Once this is working here, we can update ALL package build workflows to ignore typos for snapshot builds moving forwards.

duwudi:
> I am not convinced with the design pattern. Why should the user need to know anything about festival? Why does the user need to handle the importing and management of TTS directly?
> 
> ```
> from pitop import Pitop
> 
> pitop = Pitop()
> pitop.speak("Hello World")
> # Another option might be a 'speaker' property of the pi-top:
> # pitop.speaker.speak("Hello World")
> ```
> 
> seems like a much more sensible way to approach this.

Yeah this is exactly how it is now, though I do like the “speaker” idea as we can add more speaker-specific commands to that too (beeps, songs, pre-recorded voices etc)

m-roberts:
`pyfestival` is currently failing to build: https://github.com/pi-top/pi-topOS-Upstream-Package-Build/runs/3455683743?check_suite_focus=true

We will need to create a patch for the changes needed for this build to pass in CI.

PR #406, opened 2021-07-29
Modularise dependencies

### LINKED ISSUE #401: Separate out core SDK dependencies
A lot of the robotics stuff requires large dependencies that are not critical for working with the core components (e.g. miniscreen, battery, display).

This means that, in a Python context (particularly a virtualenv), many dependencies are required to be installed that are not explicitly required.

Separating out the dependencies like this will help to ensure that only dependencies that are required are included for a given application.

This is also a good time to consider the roll of the common library - should this simply become part of the "core" SDK?

m-roberts:
Closes #401

TODO: remove `.github/workflows/build-test-python-release.yml` (currently used to generate a wheel here: https://github.com/pi-top/pi-top-4-Miniscreen-App/tree/hypermodern/whl)

### DIFF
--- a/.github/workflows/build-test-python-release.yml
+++ b/.github/workflows/build-test-python-release.yml
@@ -1,12 +1,9 @@
-name: Deploy Test Python Release
+name: Build Test Python Release
 
 on:
   push:
     branches:
       - "rework-deps"
-  release:
-    types:
-      - published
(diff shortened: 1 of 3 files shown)


m-roberts:
Replaced by #421

PR #409, opened 2021-08-03
Vendor luma.oled and luma.core

jcapona:
We can drop all of these SDK Debian-packaged Python dependencies:
```
python3-pyftdi
python3-cbor2
python3-deprecated
python3-luma-core
python3-luma-emulator
python3-luma-oled
python3-pyusb
python3-wrapt
```

in exchange for these:
```
python3-spidev
python3-rpi.gpio
```
which are provided in the Raspberry Pi OS repository.

This means that we remove dependencies not currently available for Debian, which would need manual packaging ourselves.

There is a possible risk (low) of having to manually apply bug fixes, though `luma` is quite mature and reliable. Over time, this should evolve to become a bespoke SDK object specific for our OLED.

This does lead to a larger repo size (barely), but does lead to smaller dependency download size on install.


pitop/miniscreen/oled/core/contrib/luma/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2014-2019 Richard Hull and contributors
+# See LICENSE.rst for details.
```
angusjfw, line 3: probably we should include this MIT license

angusjfw:
Miniscreen examples still work for me.

Would be prudent to detail the tradeoffs considered in the PR. eg
- removes dependence on packages which are not currently available for Debian Bullseye, would need manual packaging ourselves
- have to manually discover and apply bug fixes
- larger repo download size
- smaller dependency download size on install

pitop/miniscreen/oled/core/contrib/luma/core/interface/serial.py
```diff
@@ -0,0 +1,189 @@
...
+
+        self._SCLK = self._configure(kwargs.get("SCLK"))
+        self._SDA = self._configure(kwargs.get("SDA"))
+        self._CE = self._configure(kwargs.get("CE"))
+        self._DC = self._configure(kwargs.get("DC"))
+        self._RST = self._configure(kwargs.get("RST"))
+        self._cmd_mode = GPIO.LOW  # Command mode = Hold low
+        self._data_mode = GPIO.HIGH  # Data mode = Pull high
+
+        if self._RST is not None:
```
angusjfw, line 64: This is probably inconsequential, but I think we are not using this RPi.GPIO bitbang implementation, and could just keep the SpiDev parts if we wanted. Haven't we had issues when using RPi.GPIO alongside gpiozero in the past? Probably best to leave this code as it was found in the luma repos for easy of merging updates though.

pitop/miniscreen/oled/core/contrib/luma/__init__.py
angusjfw: a link to where the source code was found may also prove useful, including version/commit hash

pitop/miniscreen/oled/core/contrib/luma/core/interface/serial.py
m-roberts: The reset logic highlighted here in the bitbang class is not used by the SDK, but the GPIO pin configuration is definitely still in use. `RPi.GPIO` is very much still required. The issue with it causing problems with the SDK has been resolved.

I am not expecting us to worry about merging updates, as luma's updates are rarely related to the device configuration.

PR #423, opened 2021-08-09
Add OS build info to API

m-roberts:
Move build information previously used for health check into common so that it can be used by other applications.

Currently this adds support for `/boot/pt-issue.txt` in place of `/etc/pt-issue` from buster builds, whilst aiming to provide backwards compatibility.

I've included 2 different methods that I found used for Debian version (with the same name, actually...), so it would be good to formalise this a bit better, given that it'll serve as "core application infrastructure".


pitop/common/pt_os.py
```diff
@@ -1,10 +1,44 @@
...
 from getpass import getuser
 
+
 ptissue_path = "/boot/pt-issue.txt"
 legacy_ptissue_path = "/etc/pt-issue"
 
 
+@dataclass
+class PitopOsBuildInfoV1:
+    schema_version: str
```
m-roberts, line 12: Why is `V1` in the class name if we capture the schema version?

pitop/common/pt_os.py
```diff
@@ -1,10 +1,44 @@
...
+    build_type: str
+    date: str
+    build_time_os_version: str
+    build_run_number: str
+    commit: str
+
+    @classmethod
+    def from_os_build_info(cls, build_info):
```
m-roberts, line 20: I would say that default behaviour should be to try and parse the file automatically.

pitop/common/pt_os.py
```diff
@@ -1,19 +1,74 @@
...
+        DELIMITER: str = "_"
+        LOOKUP = {
+            "S": "schema_version",
+            "B": "build_type",
+            "D": "date",
+            "P": "build_time_os_version",
+            "R": "build_run_number",
+            "#": "commit",
```
angusjfw, line 28: I think these could be more standardised, they are all based on build time, so:
date -> build_date
build_time_os_version -> build_os_version
commit -> build_commit

I know in this context it might be implied that date and commit relate to a build, and it's less clear for os_version, but thinking from the perspective of potentially passing these as parameters to the update check api it's not clear why they differ.

PR #425, opened 2021-08-10
Port type verification

m-roberts:


pitop/pma/button.py
```diff
@@ -18,6 +19,12 @@ def __init__(self, port_name, name="button"):
         self._pma_port = port_name
         self.name = name
 
+        if port_name not in Port.keys():
+            raise ValueError(f"{port_name} is not a valid port name. An example of a valid port name is D0")
+
+        if not port_name.startswith("D"):
+            raise ValueError(f"{port_name} is not a valid port type for a button. Try using a digital port, such as D0")
+
```
m-roberts, line 27: I think that this should be handled by a mixin (see `Stateful`, `Recreatable`...) so that it doesn't need to be duplicated. Not sure what I'd call it though...

duwudi: Yeah I was thinking the same, I'd just call it "AnalogComponent" or "DigitalComponent". I wouldn't necessarily call it a "mixin" though, it's just standard inheritance (`Button` _is a_ `DigitalComponent`). Admittedly the name `DigitalComponent` does kind of suggest it does more than just port checking, so maybe "mixin" is the right terminology and it should be called DigitalPortValidator 😃

m-roberts: Actually, `Button` inherits from gpiozero's `Button` class. It uses `Stateful` and `Recreatable` mixins are included, not inherited - the same would be the case for this?

duwudi: Well yeah but it's still inheritance - the only difference between a mixin and inheritance is generally just if the parent can't be used standalone it's called a "mixin". But then that definition doesn't really apply to abstract base classes, which is still inherited and definitely can't be used standalone. Anyway this is just terminology at this point, I think the idea makes sense

examples/pma/button.py
```diff
@@ -1,7 +1,7 @@
 from pitop import Button
 from time import sleep
 
-button = Button("D5")
+button = Button("D8")
```
m-roberts, line 4: I don't think this is relevant..?

pitop/core/mixins/digital_component_checks.py
```diff
@@ -0,0 +1,15 @@
...
+valid_digital_ports = ["D0", "D1", "D2", "D3", "D4", "D5", "D6", "D7"]
+
+
+class DigitalComponentChecks:
+    """Performs basic checks on validity of user-specified port."""
+
+    def __init__(self, port_name):
+        self.port_name = port_name
+
+        if port_name not in valid_digital_ports:
```
m-roberts, line 11: I think I would prefer to see regex used here to match ports: `D[0-7]`

pitop/pma/adc_base.py
```diff
@@ -22,6 +23,12 @@ class ADCBase(Stateful, Recreatable):
     """
 
     def __init__(self, port_name, pin_number=1, name="adcbase"):
+        if port_name not in Port.keys():
+            raise ValueError(f"{port_name} is not a valid port name. An example of a valid port name is A0")
```
m-roberts, line 27: It's usually better to create a new Exception class specific for things like this, but we don't really have much in the way of uniform exception handling. Perhaps this should be created as a new Issue for across the SDK rather than tackling that here.

pitop/pma/ultrasonic_sensor.py
```diff
@@ -19,10 +20,16 @@ def __init__(self,
                  name="ultrasonic"
                  ):
 
-        assert (port_name in Port)
         self._pma_port = port_name
         self.name = name
 
+        if port_name not in Port.keys():
+            raise ValueError(f"{port_name} is not a valid port name. An example of a valid port name is D0 or A1")
```
m-roberts, line 27: Same thing here re: regex and exceptions.

pitop/core/mixins/digital_component_checks.py
duwudi: good idea

pitop-wil:
@m-roberts I think this one's ready to be merged. I took into account your comments and used regex instead of dictionaries. Have tested it, and it's passed those automatic checks too. Do I just tag you here to request for it to be merged?

Issue #337, opened 2021-04-29
Improve handling of ports conflicting with in-use GPIO pins


m-roberts:
When PMA components are used on ports that map to GPIO pins that are currently in use (such as SPI0 and pin 7 from port D4), the traceback looks like this:
```
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/gpiozero/pins/native.py", line 166, in export
    result = self._exports[pin]
KeyError: 7
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/gpiozero/pins/native.py", line 177, in export
    os.O_RDONLY | os.O_NONBLOCK)
FileNotFoundError: [Errno 2] No such file or directory: '/sys/class/gpio/gpio7/value'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/pi/further/603e1ca215160f001c794b83/reference_code.py", line 13, in <module>
    ultrasonic_sensor = UltrasonicSensor("D4", threshold_distance=0.2)
  File "/usr/lib/python3/dist-packages/gpiozero/devices.py", line 124, in __call__
    self = super(GPIOMeta, cls).__call__(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/pitop/pma/ultrasonic_sensor.py", line 54, in __init__
    self.pin.edges = 'both'
  File "/usr/lib/python3/dist-packages/gpiozero/pins/__init__.py", line 420, in <lambda>
    lambda self, value: self._set_edges(value),
  File "/usr/lib/python3/dist-packages/gpiozero/pins/native.py", line 466, in _set_edges
    self.factory.fs.export(self.number)
  File "/usr/lib/python3/dist-packages/gpiozero/pins/native.py", line 181, in export
    f.write(str(pin).encode('ascii'))
OSError: [Errno 16] Device or resource busy
```

We should use a `try/except` here to catch the error and return a more meaningful error to the user if this happens.

### Proposal
**Exception text**
"Port {port_name} appears to be in use. Try using another port, or see {link_to_docs_paragraph} for more information."

**Suggestion for docs**
* Add a troubleshooting section on possible invalid pin configurations; make reference to issues with SPI0 and D4
* Add new heading to FAQ directing to troubleshooting section (Possibly as part of a broader attempt to help someone answer a question like "Why is my code suddenly not working?")

PR #427, opened 2021-08-11
Changed all D4 examples to D0

pitop-wil:
As it currently stands any code using a digital output such as a buzzer, will not work properly when connected to port D4. This is caused by the miniscreen using the same pin as port D4 (SPI 1 clash - [related issue here](https://github.com/pi-top/pi-top-Python-SDK/issues/337)). 

Unfortunately we have a lot of examples that suggest using D4 as the default port, so a lot of users will immediately run into this problem. I've changed every example port to D0 instead of D4. The underlying issue of D4 not working properly will be address in a future fix.

### DIFF
--- a/examples/pma/buzzer.py
+++ b/examples/pma/buzzer.py
@@ -1,7 +1,7 @@
 from pitop import Buzzer
 from time import sleep
 
-buzzer = Buzzer("D4")
+buzzer = Buzzer("D0")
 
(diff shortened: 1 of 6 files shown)


m-roberts (approved):
LGTM!

m-roberts:
@pitop-wil - thanks!

PR #430, opened 2021-08-12
Add miniscreen assistant class

m-roberts:
This PR creates a new `MiniscreenAssistant` class that is currently responsible for handling images (processing them for the miniscreen), and providing font paths. These will probably need to be split up and given more clear responsibilities as it grows. This class should stay undocumented until maturity.

Fetchable as a property of the class
```
from pitop import Pitop

pt = Pitop()
miniscreen = pt.miniscreen
assistant = miniscreen.assistant
```


jcapona (approved):

pitop/miniscreen/oled/assistant.py
```diff
@@ -0,0 +1,147 @@
...
+        if size < 12:
+            font_path = self.get_mono_font()
+
+        return font_path
+
+    def get_regular_font(self, size=15):
+        return ImageFont.truetype(self.get_regular_font_path(), size=size)
+
+    def get_regular_font_path(self, size=15):
+        return "Roboto-Medium.ttf"
```
jcapona, line 54:
```suggestion
        return "Roboto-Regular.ttf"
```

PR #436, opened 2021-08-31
Use new lock file path

jcapona:


m-roberts (approved):

m-roberts:
May cause issues in same session for some users updating, but this is unlikely to have severe side effects

PR #491, opened 2021-11-09
Navigation controller refactor

jcapona:
- Decouple the calculation of `linear_speed` and `angular_speed` from `NavigationController` into the `Navigator` class. These parameters are now provided in the form of a generator, and it's used by the `NavigationController` in its `__navigate` method to perform movements using `robot_move`.
- Decouple the measurement scheduler from `NavigationController` into `MeasurementScheduler`.

### DIFF
--- a/pitop/robotics/navigation/core/measurement_scheduler.py
+++ b/pitop/robotics/navigation/core/measurement_scheduler.py
@@ -1,0 +1,48 @@
+import sched
+import time
+from threading import Event, Thread
+
+
+class MeasurementScheduler:
+    def __init__(self, measurement_frequency, odometry_function, state_tracker):
+        self.state_tracker = state_tracker
+        self.odometry_func = odometry_function
+        self._measurement_dt = 1.0 / measurement_frequency
+
+        self._pose_prediction_scheduler = Thread(
+            target=self.__measurement_scheduler, daemon=True
+        )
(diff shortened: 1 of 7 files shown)


pitop/robotics/navigation/core/measurement_scheduler.py
```diff
@@ -0,0 +1,53 @@
...
+        self._new_measurement_event = Event()
+        self._measurement_prediction_scheduler = Thread(
+            target=self.__measurement_scheduler, daemon=True
+        )
+        self._measurement_prediction_scheduler.start()
+
+    def wait_for_measurement(self):
+        self._new_measurement_event.wait()
+
+    def __measurement_scheduler(self):
```
jcapona, line 28: change name

pitop/robotics/navigation/navigation_controller.py
```diff
@@ -235,12 +218,7 @@ def stop(self):
     def stop_navigation(self):
         # don't call callback if user has terminated navigation manually
         self._on_finish = None
-
-        self._stop_triggered = True
-        try:
-            self._sub_goal_nav_thread.join()
-        except Exception:
-            pass
+        self.navigator._stop_triggered = True
```
m-roberts, line 221: Not a fan of altering private member variables like this. Create a function?

PR #509, opened 2022-01-24
Fix function names and params when calling miniscreen assistant

angusjfw:


jcapona (approved):
LGTM! tested it by running the miniscreen challenges on further on my bullseye OS

PR #513, opened 2022-01-31
Add missing `distutils` dependency

m-roberts:
Used by the CLI. Unable to run `pi-top` command on Raspberry Pi OS Lite with `pi-topd` installed without recommended dependencies.

```
from distutils.util import strtobool
```
produces
```
ModuleNotFoundError: No module named 'distutils.util'
```


m-roberts:
@jcapona - can you check this? 😄

angusjfw (approved):

PR #515, opened 2022-02-28
Improve EncoderMotor class error messages

jcapona:
| Status  | Ticket/Issue |
| :---: | :--: |
| Ready| [Ticket](https://pi-top.atlassian.net/browse/SWE-10) |

#### Main changes

Error message when setting `EncoderMotor` RPMs to a value higher than maximum was confusing, since the error mentioned a maximum RPM of 6000. This is because the error is talking about the about the output of the DC motor (before the reduction gears) and not the encoder motor / wheel output.

This PR implements changes that will improve the error message by displaying the actual maximum/minimum values that the class can handle as target RPM (~114 instead of ~6000).

N/A

-

#### Manual testing tips
- Install artifacts in your pi-top[4].
- Connect the pi-top[4] to an Expansion Plate.
- Instantiate a `EncoderMotor` object and try to set the target RPM with a value higher than 114, which is the maximum:
```
>>> from pitop import EncoderMotor
>>> motor = EncoderMotor("M0", forward_direction=1)
>>> motor.set_target_rpm(200)
  File "/home/pi/pi-top-Python-SDK/pitop/pma/encoder_motor.py", line 209, in set_target_rpm
    raise ValueError(
ValueError: Target RPM value must be between -114 and 114 (inclusive)
>>> motor.set_target_rpm(144) # should start moving since the target is in range
>>> motor.set_target_rpm(-200)
  File "/home/pi/pi-top-Python-SDK/pitop/pma/encoder_motor.py", line 209, in set_target_rpm
    raise ValueError(
ValueError: Target RPM value must be between -114 and 114 (inclusive)
>>> motor.set_target_rpm(0) # should stop moving
```

-

### DIFF
--- a/pitop/pma/encoder_motor.py
+++ b/pitop/pma/encoder_motor.py
@@ -204,6 +204,12 @@ class EncoderMotor(Stateful, Recreatable):
         :param total_rotations:
             Total number of rotations to be execute. Set to 0 to run indefinitely.
         """
+
+        if not (-self.max_rpm <= abs(target_rpm) <= self.max_rpm):
+            raise ValueError(
+                f"Target RPM value must be between {-self.max_rpm} and {self.max_rpm} (inclusive)"
+            )
+
         dc_motor_rpm = int(
             round(target_rpm * self.MMK_STANDARD_GEAR_RATIO)
             * self.__forward_direction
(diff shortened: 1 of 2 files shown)


angusjfw (approved):

jcapona:
@m-roberts I think in this case it doesn't make much sense since it's a validation that's made in only one function... it's definitely a good idea for some other validations that are performed in several places

also, I just remembered that we do have a decorator-validator in the SDK that makes sure that the type of the argument that a function receives is one that the function expects (check [here](https://github.com/pi-top/pi-top-Python-SDK/blob/7422ab53c11d0df08b75f72ba471b90c567f12d2/pitop/pma/common/utils.py#L51)) - this one is used in a few places around the SDK

PR #516, opened 2022-03-01
Re-enable test CI runs

### LINKED ISSUE #414: Re-add tests
Merging `pitopcommon` to `pitop.common` (#413) breaks tests.

Instead of running in a Docker container, simply use GitHub Action Runner. Caching can be deployed for dependencies (e.g. `dlib`).

jcapona:
Closes #414

Re-enable test runs in CI using `pytest`.
Test run takes ~10 mins since it needs to build `dlib`. We could look for ways to improve this in the future (using a custom image with the dependency pre-installed or caching the dependency somehow).

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -16,3 +16,4 @@ flask_sockets
 flask_cors
 Pillow
 wget
+scipy
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,3 +1,6 @@
+pytest
+pytest-cov
+pytest-mock
 dlib
 gpiozero
 imutils
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -15,5 +15,5 @@ flask
 flask_sockets
 flask_cors
 Pillow
-wget
 scipy
+wget
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,6 +1,9 @@
+# Test framework deps
 pytest
 pytest-cov
 pytest-mock
+parameterized
+# Dependencies
 dlib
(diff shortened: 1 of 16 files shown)


m-roberts:
You should be able to cache your Python dependencies on the GitHub runner like so:
```
     - name: Install apt dependencies
       run: |
         sudo apt update
         DEBIAN_FRONTEND=noninteractive sudo apt install cmake python3-pip python3-opencv -y

     - name: Setup Python
       uses: actions/setup-python@v3.0.0
       with:
         python-version: 3.9

    - uses: actions/cache@v2.1.7
       id: cache
      with:
        path: ~/.cache/pip
        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
        restore-keys: |
          ${{ runner.os }}-pip-

    - name: Install
      if: steps.cache.outputs.cache-hit != 'true'
      run: |
        pip3 install -r tests/requirements.txt
```

You may want to run `pip3 install --upgrade pip` somewhere here as well; not sure.

m-roberts:
Also you should be able to remove `parameterized` dependency as `pytest` has [the same functionality](https://www.tutorialspoint.com/pytest/pytest_parameterizing_tests.htm).

tests/requirements.txt
```diff
@@ -9,5 +15,5 @@ flask
 flask_sockets
 flask_cors
 Pillow
-simple_pid
 wget
+scipy
```
m-roberts, line 19:
```suggestion
scipy
wget
```
alphabetical?

angusjfw (approved):
passing locally too, nice!

PR #519, opened 2022-03-21
Remove 'distutils' dependency

jcapona:
| Status  | Ticket/Issue |
| :---: | :--: |
| Ready/Hold | [Ticket](https://pi-top.atlassian.net/browse/OS-1263) |

#### Main changes

Package `distutils` isn't pip-installable, causing issues when trying to install the SDK in editable mode. 

Since `distutils` is only used to convert a string to a boolean, it's easy to replace with basic operations. My guess is that it was used for more complex command parsing but now it's only parsing the output of `raspi-config nonint` commands, which only outputs 0 or 1.

N/A

-

#### Manual testing tips
-

-

### DIFF
--- a/pitopcli/support_core/health_check.py
+++ b/pitopcli/support_core/health_check.py
@@ -1,4 +1,3 @@
-from distutils.util import strtobool
 from os import uname
 from time import strftime
 
@@ -20,7 +19,7 @@ from .ptsoftware import PitopSoftware
 
 
 def str_to_bool(value):
-    return strtobool(value) == 0
+    return str(value) == "0"
 
 
 class HealthCheck:
--- a/debian/py3dist-overrides
+++ b/debian/py3dist-overrides
@@ -1,4 +1,3 @@
-distutils python3-distutils; PEP386
 dlib python3-dlib; PEP386
 flask python3-flask; PEP386
 flask-cors python3-flask-cors; PEP386
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,8 +34,6 @@ install_requires =
     #############
     # DHCP leases
     isc_dhcp_leases >= 0.9.1
-    # CLI management
-    distutils >= 3.9.2
     # Network interfaces
     netifaces >= 0.10.4
     # Journal logging


pitopcli/support_core/health_check.py
```diff
@@ -20,7 +19,7 @@
 
 
 def str_to_bool(value):
-    return strtobool(value) == 0
+    return str(value) == "0"
```
angusjfw, line 22: this isn't a good replacement, since AFAICT this helper will be used to parse user input to questions such as 'Boot to Desktop?'. so the expected input would be eg 'y' or 'n' (although it isn't made clear to the user).

with this code "0" would be accepted as a positive response and anything else negative. just using `bool(value)` isn't good either since any non-empty string is considered positive.

[strtobool](https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool) accepts a range of common values as positive and negative responses, and throws an error if the input isn't one of those. that wouldn't be a big deal to implement here but we could probably get away with a simplification:

```suggestion
    return value.lower() in [y', 'yes', '1']
```

angusjfw (changes requested):

pitopcli/support_core/health_check.py
jcapona: The `str_to_bool(value)` function is used to parse the output of `raspi-config nonint <query>` commands. The output of the command is a `0` when the response is true and 1 otherwise.

I think you're looking at https://github.com/pi-top/pi-top-Python-SDK/blob/707a753d6d93b10585a5a5671b9d4f47f3cd3e13/pitopcli/support_core/health_check.py#L26 ... in this case, lets take for example:
```
        "get_autologin": {
            "description": "Autologin?",
            "conversion_func": str_to_bool,
        },
```

The `health-check` CLI (which is what we replaced `pt-diagnostics` with) will run `raspi-config nonint get_autologin` ... if the output is `0`, the CLI will print:
```
Autologin? : True
```

it's not parsing user responses, but the output of commands and printing them so it's easier to read the configuration of the Raspberry Pi:
```
Boot to Desktop? : True
Autologin? : True
Boot waits until a network connection is established? : False
Show splash screen at boot? : True
Filesystem can be expanded? : False
Compensation for displays with overscan? : False
Camera interface enabled? : True
SSH interface enabled? : True
VNC interface enabled? : True
SPI interface enabled? : False
I2C interface enabled? : True
Login shell accessible over serial? : True
Serial interface is enabled? : False
One-wire interface enabled? : False
GPIO server to be accessible over the network? : False
```

angusjfw: riight, thanks. I guess didn't look in detail at our code and got thrown by focusing on the `strtobool` docs.

to be fair, a function called `str_to_bool` which takes an int and inverts it's pythonic boolean meaning is a bit misleading...

angusjfw (approved):

PR #522, opened 2022-04-01
Improve SDK test coverage

jcapona:
| Status  | Ticket/Issue |
| :---: | :--: |
| Ready/Hold | [Ticket](https://pi-top.atlassian.net/browse/OS-996) |

#### Main changes
-

[insert screenshots here]

-

#### Manual testing tips
-

-


olivierwilkinson:
When running on my mac (after making a new venv and installing deps from `/tests/requirements.txt`) I get the following errors. It looks like the face and emotion detector tests need dlib and the sys info tests need certain directories to exist. I think we should add dlib to the `requirements.txt` file or mock it out. The sys info test also shouldn't be checking real directories so I think they need some better isolation.
Screenshot 2022-04-12 at 11 04 49

olivierwilkinson:
So the dlib stuff does install from the requirements.txt fine but my underlying setup is messed up from using brew for things so I think that's just a problem for my machine. I do still get one sys info test failing consistently though:
Screenshot 2022-04-14 at 10 36 52

angusjfw:
> Screenshot 2022-04-14 at 10 36 52

the helper being tested is checking for the existence of '/etc/debian_version' so clearly doesn't work outside debian... it's literally `get_debian_version`. all the helper does is read this specific file, it could only really break if the file were to be removed from debian, so I think it might be better to skip the test when not on debian, rather than mock it's assumptions entirely.

olivierwilkinson:
> > Screenshot 2022-04-14 at 10 36 52
> 
> the helper being tested is checking for the existence of '/etc/debian_version' so clearly doesn't work outside debian... it's literally called `get_debian_version`.
> 
> this could be fixed with path_mock, as in the following test, but IMO that leaves us will a less useful test.
> 
> all the helper does is read this specific file, it could only really break if the file were to be removed from debian. I think it might be better to skip the test when not on debian, rather than mock it's assumptions entirely.

Testing whether the file actually exists on the system that is running the test goes beyond what a unit test is. These tests are definitely under the unit test criteria, regardless of whether it's debatable what falls under that banner there is no structure around what the environment they are run in. I think we should ensure this test behaves as a unit test by mocking the actual check and then properly address the issue of integration / e2e testing

jcapona:
@olivierwilkinson I've added a few missing `path.exists()` mocks so it should be good now 🤞🏼

olivierwilkinson:
> @olivierwilkinson I've added a few missing `path.exists()` mocks so it should be good now 🤞🏼 

Works for me now! 🎉🙌

olivierwilkinson (approved):
LGTM

PR #527, opened 2022-04-25
Add new pre-commit checks

m-roberts:
_This ticket scraps the ticket template, as it is not public-friendly._

Found a couple of useful pre-commit checks that you might want to use (or not).

There are probably multiple files that would benefit from having their contents sorted. `debian/py3-dist-overrides` unfortunately has comments in it that are all put at the top, but you may find some use for it in another context.

Also, blocking push to master is just an obvious nice-to-have.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,6 +11,9 @@ repos:
     - id: check-yaml
     - id: check-symlinks
     - id: check-added-large-files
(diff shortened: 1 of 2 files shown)


jcapona:
hey - thanks for these, I don't think we'll add them though since we don't have a use case (yet!) for the file content sorter and the master branch is already protected!

PR #531, opened 2022-05-10
fix: update forgotten version number file

angusjfw:
| Status  | Ticket/Issue |
| :---: | :--: |
| Ready | No Ticket |

#### Main changes
- This version number, which is used for PyPi releases, has not been kept up to date. Internal issue for improving this system here: https://pi-top.atlassian.net/browse/OS-1313


jcapona (approved):
good catch!

PR #548, opened 2022-07-26
Specify camera backend on cv.VideoCapture

### LINKED ISSUE #495: Review OpenCV logs when using Camera
```java
>>> from pitop import Camera
>>>
>>> c = Camera()
[ WARN:0] global ../modules/videoio/src/cap_gstreamer.cpp (1824) handleMessage OpenCV | GStreamer warning: Embedded video playback halted; module v4l2src0 reported: Cannot identify device '/dev/video-1'.
[ WARN:0] global ../modules/videoio/src/cap_gstreamer.cpp (914) open OpenCV | GStreamer warning: unable to start pipeline
[ WARN:0] global ../modules/videoio/src/cap_gstreamer.cpp (501) isPipelinePlaying OpenCV | GStreamer warning: GStreamer: pipeline have not been created
```

The camera works fine but prints these warnings sometimes (not sure on how to consistently reproduce)

Investigate what causes these errors, presumably it’s some part of [https://github.com/pi-top/pi-top-Python-SDK/blob/45fa20fa93d4a5bd22038f41d6f4da0d06d675dc/pitop/camera/core/cameras/usb_camera.py](https://github.com/pi-top/pi-top-Python-SDK/blob/45fa20fa93d4a5bd22038f41d6f4da0d06d675dc/pitop/camera/core/cameras/usb_camera.py|smart-link) . Could it be related to searching for the correct device by trial and error? We don’t do this explicitly anymore but maybe cv.VideoCapture does it internally when we pass -1. but i wasn’t able to repro by trying different ports

Can we avoid this being triggered, or somehow suppress the warnings - this should be possible from python?

some relevant search results:

[https://stackoverflow.com/questions/70624679/gstreamer-warning-when-running-opencv-on-ubuntu](https://stackoverflow.com/questions/70624679/gstreamer-warning-when-running-opencv-on-ubuntu|smart-link) 

[https://community.intel.com/t5/Intel-Distribution-of-OpenVINO/GStreamer-Error-when-initiating-OpenCV-with-filename-as-cv2/m-p/1361963](https://community.intel.com/t5/Intel-Distribution-of-OpenVINO/GStreamer-Error-when-initiating-OpenCV-with-filename-as-cv2/m-p/1361963|smart-link)

jcapona:
| Status  | Ticket/Issue |
| :---: | :--: |
| Ready/Hold | [Ticket](https://pi-top.atlassian.net/browse/OS-1173) |

Closes  #495 

#### Main changes

Specify the `V4L2` backend for the USB camera capture object, which tells the video capture object that the cameras support capturing via `libv4l`. The default value `CAP_ANY` would probe the device to determine the correct backend to use, one of them being `CAP_GSTREAMER`, which prints the warning messages when unable to initialize.

The list of available backends can be found [here](https://docs.opencv.org/4.x/d4/d15/group__videoio__flags__base.html#ga023786be1ee68a9105bf2e48c700294d). Also, [this](https://docs.opencv.org/4.x/d0/da7/videoio_overview.html) is an interesting read. 

[insert screenshots here]

-

#### Manual testing tips
-

-

### DIFF
--- a/pitop/camera/core/cameras/usb_camera.py
+++ b/pitop/camera/core/cameras/usb_camera.py
@@ -33,7 +33,8 @@ class UsbCamera:
             self._rotate_angle = rotate_angle
 
         def create_camera_object(index, resolution=None):
-            cap = import_opencv().VideoCapture(index)
+            cv = import_opencv()
+            cap = cv.VideoCapture(index, cv.CAP_V4L2)
             if resolution is not None:
                 cap.set(3, resolution[0])
                 cap.set(4, resolution[1])
--- a/pitop/camera/core/cameras/usb_camera.py
+++ b/pitop/camera/core/cameras/usb_camera.py
@@ -34,7 +34,18 @@ class UsbCamera:
 
         def create_camera_object(index, resolution=None):
             cv = import_opencv()
-            cap = cv.VideoCapture(index, cv.CAP_V4L2)
+            try:
+                # Force V4L2 backend first to avoid warning logs
+                cap = cv.VideoCapture(index, cv.CAP_V4L2)
+            except Exception:
+                cap = None
+
+            # Retry using default backend on failure.
+            # Check if device is opened since sometimes VideoCapture
+            # doesn't raise an Exception on error
+            if cap is None or not cap.isOpened():
+                cap = cv.VideoCapture(index)
+
             if resolution is not None:
                 cap.set(3, resolution[0])
                 cap.set(4, resolution[1])


angusjfw (commented):
Cool, nice find. A few things to consider:

- can we test that this has removed the warning log - i was not able to reproduce them
- presumably this limits the cameras which this class supports
  - can we document what is and isn't supported and point to alternatives. is this usb & linux only? rpi camera port?
  - are we sure we don't need to support other backends - we don't already use this for cv images or ffmpeg steams somewhere in our robotics or something? 
  - if needed can we support others. we could add another camera type or maybe try catch here ourselves and try the any as a backup.

jcapona:
> Cool, nice find. A few things to consider:
> 
> * can we test that this has removed the warning log - i was not able to reproduce them

This removed the warning logs for me and I was getting them consistently. Since we don't know a way to trigger the warning messages, I don't think there's an 100% effective way to test this other than not getting the logs.

If you create the `VideoCapture` object using the `cv.CAP_GSTREAMER` backend, you'll get the warning:
```
>>> import cv2
>>> c = cv2.VideoCapture(0, cv2.CAP_GSTREAMER)
[ WARN:0] global ../modules/videoio/src/cap_gstreamer.cpp (1824) handleMessage OpenCV | GStreamer warning: Embedded video playback halted; module v4l2src0 reported: Internal data stream error.
[ WARN:0] global ../modules/videoio/src/cap_gstreamer.cpp (914) open OpenCV | GStreamer warning: unable to start pipeline
[ WARN:0] global ../modules/videoio/src/cap_gstreamer.cpp (501) isPipelinePlaying OpenCV | GStreamer warning: GStreamer: pipeline have not been created
```   
Which I guess is what happens when opencv probes for devices/backends. If you run the same code as before but using `cv2.CAP_V4L`, you won't see those warnings.

> * presumably this limits the cameras which this class supports
>   
>   * can we document what is and isn't supported and point to alternatives. is this usb & linux only? rpi camera port?

Yes, V4L only works in linux systems and should work for any camera that supports it and creates a video device in `/dev/` (such as the RPi camera or other USB cameras). 

>   * are we sure we don't need to support other backends - we don't already use this for cv images or ffmpeg steams somewhere in our robotics or something?

The SDK doesn't use any other backend; the only thing we're doing here is avoid probing by passing the expected backend directly. We know this is the particular backend used by our camera since it's what's returned by `getBackendName`:
```
>>> import cv2
>>> c = cv2.VideoCapture(-1)
[ WARN:0] global ../modules/videoio/src/cap_gstreamer.cpp (1824) handleMessage OpenCV | GStreamer warning: Embedded video playback halted; module v4l2src0 reported: Cannot identify device '/dev/video-1'.
[ WARN:0] global ../modules/videoio/src/cap_gstreamer.cpp (914) open OpenCV | GStreamer warning: unable to start pipeline
[ WARN:0] global ../modules/videoio/src/cap_gstreamer.cpp (501) isPipelinePlaying OpenCV | GStreamer warning: GStreamer: pipeline have not been created
>>> c.getBackendName()
'V4L2'
``` 
The file system camera handles images with `PIL` and the robotics/algorithms don't care about this. I tested this change with the SDK examples and they continue to work (e.g.: face detector, emotion detector, line detection, ...)

>   * if needed can we support others. we could add another camera type or maybe try catch here ourselves and try the any as a backup.

Yeah, good idea. I just commited a change to fallback to the old code (with no specified backend) on failure, which should cover us in the case where something unexpected happens.

angusjfw (approved):

PR #557, opened 2022-09-21
Depend on numpy>1.19.5 when pip-installing

jcapona:
#### Main changes
When pip installing a subpackage in Raspberry Pi OS, we used the already installed `numpy` package, version `1.19.5`, which was the version we wanted to install. However, this error came up when trying to import `cv2`:
```
>>> import cv2
RuntimeError: module compiled against API version 0xf but this version of numpy is 0xd
  File "/home/pi/.local/lib/python3.9/site-packages/cv2/__init__.py", line 8, in <module>
    from .cv2 import *
ImportError: numpy.core.multiarray failed to import
```
After reading this [useful explanation](https://stackoverflow.com/a/49756785), it seems that we need to use a newer version of  `numpy`. 

Note: This is not an issue when installing the subpackages using the `deb` files.

[insert screenshots here]

-

#### Manual testing tips
-

-

### DIFF
--- a/packages/camera/setup.py
+++ b/packages/camera/setup.py
@@ -46,9 +46,9 @@ __requires__ = [
     f"pitop.pma=={__version__}",
     f"pitop.processing=={__version__}",
     "imageio>=2.4.1,<3.0.0",
-    "numpy>=1.19.5,<2.0.0",
+    "numpy>1.19.5,<2.0.0",
     "Pillow>=8.1.2,<9.0.0",
-    "opencv-python>=4.5.1,<5.0.0",
+    "opencv-python>=4.5.1,<4.6.0",
 ]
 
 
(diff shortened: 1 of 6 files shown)


jcapona:
Yeah, on the deb side, we're pretty much stuck with `numpy 1.19.5` and `opencv 4.5.1`, and both of them work together okay.

If you pip install the same versions, sadly, they're not compatible with each other (`module compiled against API version 0xf but this version of numpy is 0xd`). My guess is that the version of `opencv` distributed via pypi was compiled using a different version of `numpy` compared to the one distributed via `apt`.

The main issue we need to solve here is that we need to make sure that the version of `numpy` is `>1.19.5`, at least when pip-installing. The limitation on the version of `opencv` is not really necessary, since the latest version available in pypi, `4.6.0`, is still compatible with the version of `numpy` that gets installed via pip, `1.23.0`.

angusjfw (approved):

PR #558, opened 2022-09-21
Add switch_user common helpers

angusjfw:
| Status  | Ticket/Issue |
| :---: | :--: |

#### Main changes
- Adds helpers for running process as another user, taken from further-link. This is intended to be shared between further-link and the miniscreen's new [project menu](https://github.com/pi-top/pi-top-4-Miniscreen/pull/283/files).

[insert screenshots here]

- I've tested manually but it could do with a test suite.

#### Manual testing tips
-

-

### DIFF
--- a/pitop/common/switch_user.py
+++ b/pitop/common/switch_user.py
@@ -1,14 +1,16 @@
+import grp
 import os
 import pwd
-import grp
 
 
 def env_for_user(user, env=None):
     """Returns an environment variables dictionary, intended for use as the env
-    parameter when spawining a process as another user. Takes the initial env
-    as a param or from the current environment and modifies user-specific parts
-    to suit the new user. Use-case specific variables such as DISPLAY should be
-    set outside."""
+    parameter when spawining a process as another user.
+
+    Takes the initial env as a param or from the current environment and
+    modifies user-specific parts to suit the new user. Use-case specific
+    variables such as DISPLAY should be set outside.
+    """
 
     if get_uid(user) is None:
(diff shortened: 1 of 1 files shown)


jcapona:
I think we should probably move the `get_*` functions into the `pitop.common.current_session_info` file.
Also, should our `pitop.command_runner.run` function use the user environment from this PR?

angusjfw:
> I think we should probably move the `get_*` functions into the `pitop.common.current_session_info` file.

Even though these functions resemble some in there, like get_current_user, they are actually not related to a 'current session' - you have to provide a user argument and they provide fairly static information about the system users. They could have a more common/generic use than being internals of 'switch_user.py', but I'd personally suggest leaving them there until this use case emerges. They're mostly thin wrappers on the pwd module anyway. Maybe I should prefix them an underscore or two?

> Also, should our `pitop.command_runner.run` function use the user environment from this PR

That isn't necessary unless we want to use command_runner as a different user, in which case we could add a param or separate method to command_runner and make use of these. The intention is that you could use `env_for_user` alongisde other use-case-specific environment changes like we have in command_runner. So if we did it might look like:
```
def run_command_background(command_str: str, print_output=False, user=None) -> Popen:
    if user is not None:
      return Popen(
          split(command_str),
          env=env_for_user(user, __get_env()),  # combining the switch_user env with command_runner's get_env-
          preexec_fn=lambda: switch_user(user)
...
```

jcapona (approved):

PR #559, opened 2022-09-23
Add functions to manage sections in config files

jcapona:
| Status  | Ticket/Issue |
| :---: | :--: |
| Ready/Hold | [Ticket](https://pi-top.atlassian.net/browse/OS-1369) |

#### Main changes
Add functions to add/remove sections from configuration files. The convention used for the blocks to add/remove from files is as follows:
 ```
# [<TITLE>] START
#
# <DESCRIPTION>
#
# Notes:
# <NOTES>
#
<CONTENT>
#
# [<TITLE>] END
```
Which is similar to what we already have in some configuration files in pi-topOS.

[insert screenshots here]

-

#### Manual testing tips
-

-

### CODE REVIEW

File: pitop/common/configuration_file.py [hunk 1]
```diff
@@ -0,0 +1,86 @@
...
+        #
+        # [<APP_NAME>] END
+    """
+    start_pattern = compile(rf"^(# \[{title}\] START)\n")
+    finish_pattern = compile(rf"^(# \[{title}\] END)\n")
+    delete_line = False
+
+    with open(filename, "r") as fr:
+        lines = fr.readlines()
+        with open(filename, "w") as fw:
```
angusjfw, line 61: is it really necessary to have separate read & write handles? you can't use `with open(filename, "rw") as file:`?

pitop/common/configuration_file.py
```diff
@@ -0,0 +1,86 @@
...
+            for i, line in enumerate(lines):
+                if (
+                    line.strip() == ""
```
angusjfw, line 64: why should the line before the start pattern have to be empty? when you create the section you don't add an empty line before it?

jcapona: yeah, the string is:
```
text = f"""
# ...
```
which counts as a new line

pitop/common/configuration_file.py
jcapona: updated! I tried this before, but it just didn't work; adding `f.truncate()` after all the `write()` calls did it for me

pitop/common/configuration_file.py
angusjfw: right. it might still be better to drop the requirement for this empty line. what if someone edits the file manually and removes it, it doesn't look important. 

as they say, if `start_pattern.search(lines[i + 1]) is not None`, who cares whether `line.strip() == ""`

jcapona: makes sense - just updated it!

angusjfw: tbh i thought you could remove the check for an empty line since it's non-essential, but leave the actual line just for better formatting

angusjfw (approved):

angusjfw (approved):

PR #578, opened 2022-10-17
Simulate miniscreen

angusjfw:
| Status  | Ticket/Issue |
| :---: | :--: |
| Ready | [Ticket](https://pi-top.atlassian.net/browse/OS-1355) |


packages/simulation/pitop/simulation/virtual_hardware/pitop.py
```diff
@@ -0,0 +1,55 @@
+from unittest.mock import Mock as TMock
```
jcapona, line 1: why not just `Mock`?

examples/simulation/pitop_simulate.py
```diff
@@ -29,18 +27,17 @@
...
 
-pitop_sim = simulate(pitop)
+pitop_sim = simulate(pitop, 0.5)
 led_sim = simulate(pitop.led4)
 button_sim = simulate(pitop.button4)
 
-import pygame  # noqa: E402
+from os import path  # noqa: E402
+
+from PIL import Image  # noqa: E402
```
jcapona, line 36: I think these should go to the top of the file and remove the `noqa` comments

examples/simulation/pitop_simulate.py
```diff
@@ -2,8 +2,6 @@
 
 use_virtual_hardware()
 
-from time import sleep  # noqa: E402
-
 from pitop import Pitop  # noqa: E402
 from pitop.pma import LED, Button  # noqa: E402
```
jcapona, line 6: I think these should go to the top of the file... unless `use_virtual_hardware` is needed at the very top of the file?

jcapona (approved):
great stuff! Left some questions/comments but don't think they're blockers

examples/simulation/pitop_simulate.py
angusjfw: yeah component imports have to come after `use_virtual_hardware` so that they use the mocks - i've added a comment at least. users aren't really meant to have call `use_virtual_hardware` so this shouldn't cause friction normally (they'd use real hardware or it's mocked during import on a cloud server with the env var PITOP_VIRTUAL_HARDWARE)

jcapona (approved):

PR #590, opened 2022-11-11
PTDM subscribes to relevant events only

jcapona:
#### Main changes

Improve `PTDMSubscribeClient` structure by:
- Only subscribe to the events that were marked as relevant according to the dictionary passed to the `initialise` method. By doing this, overhead is reduced since it's not necessary to filter the received messages once they're received since now when that happens, we're 100% sure it's from an event we're interested in.
- Stop using `zmq.Poller`, since it's not necessary. The `recv_string` method will unblock when a valid message is received.

[insert screenshots here]

-

#### Manual testing tips
-

-

### DIFF
--- a/packages/common/pitop/common/ptdm.py
+++ b/packages/common/pitop/common/ptdm.py
@@ -414,7 +414,7 @@ class PTDMSubscribeClient:
     def __init__(self):
         self.__thread = Thread(target=self.__thread_method, daemon=True)
 
-        self._callback_funcs = None
+        self._callback_funcs = {}
 
         self._zmq_context = None
         self._zmq_socket = None
@@ -429,7 +429,9 @@ class PTDMSubscribeClient:
     def __connect_to_socket(self):
         self._zmq_context = zmq.Context()
         self._zmq_socket = self._zmq_context.socket(zmq.SUB)
-        self._zmq_socket.setsockopt_string(zmq.SUBSCRIBE, "")
+
+        for message_id in self._callback_funcs.keys():
+            self._zmq_socket.setsockopt(zmq.SUBSCRIBE, str(message_id).encode())
 
         try:
             self._zmq_socket.connect(self.URI)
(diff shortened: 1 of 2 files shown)


packages/common/pitop/common/ptdm.py
```diff
@@ -457,35 +459,37 @@ def __thread_method(self):
...
-        if func_arg_no > 1:
-            logger.error(
-                "Invalid callback function - it should receive at most one argument."
-            )
-            return ""
+                callback = self._callback_funcs.get(message.message_id())
+                if callback:
+                    self.invoke_callback_func_if_exists(callback, message.parameters)
 
+    def invoke_callback_func_if_exists(self, callback, params=list()):
```
angusjfw, line 470: i'm not sure but this looks like it would have the same issue as using `[]` as a default, that the same list is actually reused each time. probably not an issue since you don't mutate it, but I still think it's better practice to use None as the default and then have like `params = list() if params is None else params`

https://nikos7am.com/posts/mutable-default-arguments/

angusjfw: also, not new things but:
- the naming of this function is confusing me - it doesn't check wether the callback exists it just calls it?
- below is `if params == list()` just a weird way to check if the list is empty? usually people check the length?

packages/common/pitop/common/ptdm.py
```diff
@@ -457,35 +459,37 @@ def __thread_method(self):
...
+        func_arg_no = len(signature(callback).parameters)
         if params == list() or func_arg_no == 0:
-            func()
+            callback()
         else:
-            func(params)
+            callback(params)
```
angusjfw, line 475: these lines aren't covered by tests - maybe you could quickly add one where the callback is actually called?

angusjfw (commented):
the message filtering thing seems to make sense and be correct to me, but it seems like you have removed parts of the test that actually check the callbacks work

PR #594, opened 2022-11-24
Simulate PMA components

jcapona:
| Status  | Ticket/Issue |
| :---: | :--: |
| Ready/Hold | [Ticket](https://pi-top.atlassian.net/browse/OS-1354) |

#### Main changes

Add missing PMA components to sprites using a pygame slider to display/control component values.

[insert screenshots here]

-

#### Manual testing tips
-

-


angusjfw (commented):
The crop on these new PMA cube images is not quite as tight as the ultrasonic, buzzer, led. Being super picky here but we can take another pixel or two of background off them and then resize again to make sure they all appear the same size.

angusjfw (commented):
We need to take the sim scale into account throughout the new sprites. All the images are being scaled but other bits like the base image and sliders are not. I'll add comments to some of the areas that need to consider scaling.

At the moment if you modify the sensors_simulate.py example to use `simulate(pitop, 0.5)` you get something like this where the sensors have much larger white borders than necessary which overlap the pi-top:

image

and if you scale up by 2, the base images are not large enough for their contents:
image

In addition to scaling we could try making the base images transparent to prevent this corner overlap issue, although I had issues before when using transparency that the rendering was not consistent and so snapshot testing was difficult (this is why the _load_image helper adds a white background.)

We might also need to reconsider how scale is used in the _generate_sprite_centres, although I think it is accounted for at least partially by using the pre-scaled sim_size and  main_sprite_rect.

packages/simulation/pitop/simulation/sprites.py
```diff
@@ -85,3 +102,421 @@ def handle_event(type, target_name, component):
...
+                int((self.Size[1] - image.get_height()) / 2),
+            )
+
+        self.component_image = SimSprite._load_image(Images.Buzzer, scale)
+        self.component_pos = get_pos(self.component_image)
+
+        self.buzzer_sound_image = SimSprite._load_image(Images.buzzer_sound_icon, scale)
+        self.buzzer_sound_image_pos = get_pos(self.buzzer_sound_image)
+
+        self.base_image = pygame.Surface(self.Size)
```
angusjfw, line 125: this is not accounting for the sim scale. when these Size properties are used for the overall sim size they are [scaled](https://github.com/pi-top/pi-top-Python-SDK/blob/e436c14cb53a7d50db511ca8b01e90744ef3903c/packages/simulation/pitop/simulation/simulation.py#L128).

packages/simulation/pitop/simulation/sprites.py
```diff
@@ -85,3 +102,421 @@ def handle_event(type, target_name, component):
...
+
+    def update(self):
+        if not hasattr(self, "state"):
+            return
+
+        self.image = self.render()
+
+
+class LightSensor(pygame.sprite.Sprite, SimSprite):
+    Size = ((PMA_CUBE_SIZE[0] + 2 * MARGIN) * 2, PMA_CUBE_SIZE[1] + 2 * MARGIN)
```
angusjfw, line 149: this is used repeatedly and a bit hard to read so maybe use a constant like
```
from .simsprite import PMA_CUBE_SIM_SIZE
DOUBLE_WIDTH_PMA_CUBE_SIM_SIZE = (PMA_CUBE_SIM_SIZE[0] * 2, PMA_CUBE_SIM_SIZE[1])
```

packages/simulation/pitop/simulation/sprites.py
angusjfw: it looks like this size is used quite a bit in positioning later so probably we should save the scaled size onto the instance.

eg self.size = [scale * x for x in self.Size]

might want to rename these to class.SimSize and instance.sprite_size or something like that to make it clearer

the Size class properties were mainly for determining the overall window size if this sprite is the main sprite, since in cases like the Pitop this isn't directly related to the size of the pitop sprite itself

packages/simulation/pitop/simulation/sprites.py
```diff
@@ -85,3 +102,421 @@ def handle_event(type, target_name, component):
...
+
+    def __init__(self, config, scale):
+        pygame.sprite.Sprite.__init__(self)
+
+        self.base_image = self.get_base_image(scale)
+
+        lightbulb_image = SimSprite._load_image(Images.lightbulb_icon, scale)
+        lightbulb_pos = (
+            int(self.Size[0] / 4 * 3 - lightbulb_image.get_width() / 2),
```
angusjfw, line 158: size here doesn't account for scaling.

packages/simulation/pitop/simulation/sprites.py
```diff
@@ -85,3 +102,421 @@ def handle_event(type, target_name, component):
...
+        if not is_virtual_hardware():
+            self.slider.disable()
+
+        self.image = self.render()
+        self.rect = self.image.get_rect()
+
+    def get_base_image(self, scale):
+        component_image = SimSprite._load_image(Images.LightSensor, scale)
+
+        base_image = pygame.Surface(self.Size)
```
angusjfw, line 176: no scaling here

packages/simulation/pitop/simulation/sprites.py
```diff
@@ -85,3 +102,421 @@ def handle_event(type, target_name, component):
...
+        base_image.fill(pygame.Color("WHITE"))
+        base_image.blit(component_image, (MARGIN, MARGIN))
```
angusjfw, line 178: i suppose even the margins need to be scaled otherwise things wont be positioned correctly relative to the base_image

angusjfw (commented):
it would be good to add one of these more complex components into the test `test_pitop_simulate_scale` so that we have tested their scaling behaviour a bit. you'd might need to adjust the simulate 'size' parameter in that test to ensure the new component is included in the snapshot

packages/simulation/pitop/simulation/sprites.py
```diff
@@ -85,3 +102,421 @@ def handle_event(type, target_name, component):
...
+        )
+
+    def render(self):
+        value = 0
+        if hasattr(self, "state"):
+            value = self.state.get("reading", 0)
+
+        image = self.base_image.copy()
+        image.blit(
+            self.font.render(str(value), True, pygame.Color("BLACK")), self.text_pos
```
angusjfw, line 210: can we replace the black and grey of the text and sliders with the shades from our colour scheme which will match the icons:
#505050 instead of black (text & slider ball)
#E6E8EA for grey (slider bar) or maybe #DBDEE0 if more contrast is needed

packages/simulation/pitop/simulation/sprites.py
```diff
@@ -85,3 +100,233 @@ def handle_event(type, target_name, component):
...
+        self.slider = Slider(
+            x=slider_pos[0],
+            y=slider_pos[1],
+            width=int(SLIDER_WIDTH * self.scale),
+            height=int(SLIDER_HEIGHT * self.scale),
+            min=self.min_value,
+            max=self.max_value,
+            step=self.slider_step,
+            initial=0,
+            handleRadius=int(SLIDER_HANDLE_RADIUS * self.scale**2),
```
angusjfw, line 201: I don't think this needs to be squared

angusjfw (approved):

PR #595, opened 2022-11-28
Set 'number_of_samples' in constructor of ADCBase and subclasses

jcapona:
| Status  | Ticket/Issue |
| :---: | :--: |
| Ready/Hold | [Ticket](https://pi-top.atlassian.net/browse/OS-1393) |

#### Main changes
- Set 'number_of_samples' in constructor of ADCBase and subclasses
- Add range of returned value in SoundSensor

[insert screenshots here]

-

#### Manual testing tips
-

-

### DIFF
--- a/packages/pma/pitop/pma/adc_base.py
+++ b/packages/pma/pitop/pma/adc_base.py
@@ -18,17 +18,23 @@ class ADCBase(Stateful, Recreatable):
     :param str port_name: The ID for the port to which this component is connected
     """
 
-    def __init__(self, port_name, pin_number=1, name="adcbase"):
+    def __init__(self, port_name, pin_number=1, name="adcbase", number_of_samples=1):
         self._pma_port = port_name
         self.name = name
 
-        self.is_current = False
+        self.number_of_samples = number_of_samples
         self.channel = get_pin_for_port(self._pma_port, pin_number)
         self.__adc_device = PlateInterface().get_device_mcu()
 
         Stateful.__init__(self)
         Recreatable.__init__(
-            self, {"port_name": port_name, "pin_number": pin_number, "name": self.name}
+            self,
+            {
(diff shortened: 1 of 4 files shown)


packages/pma/pitop/pma/adc_base.py
```diff
@@ -38,7 +44,7 @@ def own_state(self):
         }
 
     def read(
-        self, number_of_samples=1, delay_between_samples=0.05, peak_detection=False
+        self, delay_between_samples=0.05, peak_detection=False, number_of_samples=None
```
angusjfw, line 47: is there a reason for reordering these params? i think that can be considered a breaking change in python since keyword params can still be used positionally?

packages/pma/pitop/pma/light_sensor.py
```diff
@@ -15,8 +15,16 @@ class LightSensor(ADCBase):
     :param str port_name: The ID for the port to which this component is connected
     """
 
-    def __init__(self, port_name, pin_number=1, name="light_sensor"):
-        ADCBase.__init__(self, port_name=port_name, pin_number=pin_number, name=name)
+    def __init__(
+        self, port_name, pin_number=1, name="light_sensor", number_of_samples=3
```
angusjfw, line 19: i think we should mention in the doc description that by default 3 samples are taken so the reading takes around 0.5 seconds

packages/pma/pitop/pma/adc_base.py
jcapona: my bad!

packages/pma/pitop/pma/light_sensor.py
jcapona: makes sense; also just added documentation for the `name` parameter which was missing

angusjfw (approved):

PR #598, opened 2022-12-16
Display port label on top of components when simulating

jcapona:
| Status  | Ticket/Issue |
| :---: | :--: |
| Ready/Hold | [Ticket](https://pi-top.atlassian.net/browse/OS-1357) |

#### Main changes
- Display port label on top of components when they're part of a `Componentable` using a new `PortLabel` component.
- Remove margin from sensors and buzzer.

[insert screenshots here]

-

#### Manual testing tips
-

-


packages/simulation/pitop/simulation/port_label.py
```diff
@@ -0,0 +1,55 @@
...
+    def size(self):
+        return multiply_scalar(self.scale, (self.width, self.height))
+
+    @property
+    def color(self):
+        if isinstance(self.port_name, str) and self.port_name[0] == "A":
+            return pygame.Color("#008BCD")
+        if isinstance(self.port_name, str) and self.port_name[0] == "D":
+            return pygame.Color("#DF1995")
+        return pygame.Color("2B2B2B")
```
angusjfw, line 48: looks like this default is missing the # symbol
```suggestion
        return pygame.Color("#2B2B2B")
```

it could be a good time to refactor the colours into constants or an enum

angusjfw: i notice you've added  PITOP_GREY_COLOR so even more reason to get it standardised

angusjfw (commented):
These look great.

I like how you've added it as a 'component' rendering over a provided image. The parallels with the miniscreen component library are interesting. Maybe in future we will formalise this a bit more like we did there but it's good enough for now.

Maybe you could implement the draw_port stuff generically on the SimSprite class somehow... you've managed to achieve it pretty well in the SliderSensorSprite... could you move that up to a SimSprite super constructor maybe? That could lead us to a more standardised 'Component' system where they're all based on base_images and render methods rather than the ad-hoc stuff we've got in the simpler ones... This would be great to have a go at but we could create a ticket for later if you don't want to approach it now.

packages/simulation/pitop/simulation/sprites.py
```diff
@@ -199,7 +252,7 @@ def create_components(self):
             step=self.slider_step,
             initial=0,
             handleRadius=int(SLIDER_HANDLE_RADIUS * self.scale),
-            handleColour=pygame.Color("#505050"),
+            handleColour=PitopColor.CHARCOAL.value,
```
angusjfw, line 255:
```suggestion
            handleColour=pygame.Color(PitopColor.CHARCOAL.value),
```

This is actually the reason for the broken/stalling tests. Pytest and xdist make this not at all clear to debug, but you can see the error if you use pytest -s flag (or probably also if you disable xdist):
```
venv ❯ pytest tests/test_pitop.py -s -k test_pitop_simulate
...blahblah
Process Process-1:
  File "/usr/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/lib/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/angus/pi-top/src/pi-top-Python-SDK/packages/simulation/pitop/simulation/simulation.py", line 148, in _run
    sprite_group = sprite_class.create_sprite_group(size, config, scale)
  File "/home/angus/pi-top/src/pi-top-Python-SDK/packages/simulation/pitop/simulation/simsprite.py", line 83, in create_sprite_group
    sprite = sprite_class(component, scale, draw_port=True)
  File "/home/angus/pi-top/src/pi-top-Python-SDK/packages/simulation/pitop/simulation/sprites.py", line 230, in __init__
    self.image = self.render()
  File "/home/angus/pi-top/src/pi-top-Python-SDK/packages/simulation/pitop/simulation/sprites.py", line 313, in render
    self.slider.draw(image)
  File "/home/angus/pi-top/src/pi-top-Python-SDK/packages/simulation/pitop/simulation/widgets/slider.py", line 123, in draw
    gfxdraw.filled_circle(
TypeError: invalid color argument
```

packages/simulation/pitop/simulation/sprites.py
```diff
@@ -199,7 +252,7 @@ def create_components(self):
...
             colour=pygame.Color("#DBDEE0"),
```
angusjfw, line 256: we should use the enum for this one too, though I wish they didn't need the .value thing

packages/simulation/pitop/simulation/sprites.py
```diff
@@ -240,14 +303,15 @@ def render(self):
 
         image = self.base_image.copy()
         image.blit(
-            self.font.obj.render(str(value), True, pygame.Color("#505050")),
+            self.font.obj.render(str(value), True, PitopColor.CHARCOAL.value),
```
angusjfw, line 306: probably we need should do pygame.Color in other places like this, though it doesn't seem to be actually broken here

angusjfw (approved):

PR #616, opened 2023-06-23
Handle OLED & button locks in Miniscreen class

jcapona:


angusjfw (approved):
have not tested but looks good

PR #620, opened 2023-07-25
Fix test runs 

jcapona:
- Updated oled/miniscreen mocks; some tests where hanging because of a locked lock.
- Converted some tests from unittest to pytest
- Added `daemon=True` to all `Threads`
- Updated `Mock` to `MagicMock`
- Added `simulationtest` test group to pytest. Now we can run these by using `-m` when running `pytest` (eg: `pytest -vvvv -m simulationtest`

On test workflow:
- Added a timeout of 300s to kill tests that hang (Using `pytest-timeout`)
- Run simulation tests without an actual display (by setting the `SDL_VIDEODRIVER` env var)


angusjfw (approved):
Nice. The simulation snapshots weren't quite consistent for me locally but at least they are here

PR #665, opened 2024-09-17
Create camera using all indexes if auto detection fails

jcapona:
| Status  | Ticket/Issue |
| :---: | :--: |

#### Main changes
-

[insert screenshots here]

-

#### Manual testing tips
-

-

### DIFF
--- a/packages/camera/pitop/camera/core/cameras/usb_camera.py
+++ b/packages/camera/pitop/camera/core/cameras/usb_camera.py
@@ -57,7 +57,7 @@ class UsbCamera:
             return cap
 
         self._camera = create_camera_object(index, resolution)
-        if self._camera is None and index is None:
+        if self._camera is None and index is -1:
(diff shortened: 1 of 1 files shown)


packages/camera/pitop/camera/core/cameras/usb_camera.py
```diff
@@ -46,12 +51,19 @@ def create_camera_object(index, resolution=None):
...
 
-            if resolution is not None:
+            if cap is not None and resolution is not None:
                 cap.set(3, resolution[0])
                 cap.set(4, resolution[1])
             return cap
 
-        self._camera = create_camera_object(self.index, resolution)
+        self._camera = create_camera_object(index, resolution)
+        if self._camera is None and index is None:
```
angusjfw, line 60:
```suggestion
        if self._camera is None and index is -1:
```
we only want to scan if the index was not specified by the user - which would mean it is now -1

PR #687, opened 2025-03-12
DriveController set synchronous mode on each movement

jcapona:
| Status  | Ticket/Issue |
| :---: | :--: |

#### Main changes

The DriveController enables synchronous mode in its constructor, which causes issues when trying to use an EncoderMotor individually; this PR will enable synchronous mode for each movement of the DriveController and then disable it, allowing to use an EncoderMotor alongside the DriveController.

Sample snippet:

```
from pitop import DriveController

d = DriveController()
d.forward(0.5)
d.stop()

left_motor = d.left_motor

# this doesn't work right now
left_motor.set_target_rpm(144)
```

[insert screenshots here]

-

-

### DIFF
--- a/packages/robotics/pitop/robotics/drive_controller.py
+++ b/packages/robotics/pitop/robotics/drive_controller.py
@@ -60,7 +60,6 @@ class DriveController(Stateful, Recreatable):
 
         # Motor syncing
         self.__mcu_device = PlateInterface().get_device_mcu()
-        self._set_synchronous_motor_movement_mode()
 
         Stateful.__init__(self, children=["left_motor", "right_motor"])
         Recreatable.__init__(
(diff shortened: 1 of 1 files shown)


packages/robotics/pitop/robotics/drive_controller.py
```diff
@@ -79,8 +78,13 @@ def _set_synchronous_motor_movement_mode(self) -> None:
         )
         self.__mcu_device.write_byte(MotorSyncRegisters.CONFIG.value, sync_config)
 
+    def _unset_synchronous_motor_movement_mode(self) -> None:
+        self.__mcu_device.write_byte(MotorSyncRegisters.CONFIG.value, 0b0000000)
+
     def _start_synchronous_motor_movement(self) -> None:
+        self._set_synchronous_motor_movement_mode()
```
angusjfw, line 85: instead of doing this sync mode either side of the 'start' we need to do it either side of setting the target rpms and start, eg doing it and the start and end of  _set_motor_speeds(

angusjfw (approved):

